PR #24, opened 2017-02-05
Developed initial implementation for IDtoModelMapperConfiguration

develqi:
It ignores properties based on https://github.com/AutoMapper/AutoMapper/issues/790 automatically + Sets CreateMissingTypeMaps to true by default + Removed its equivalent code from IDependencyManangerExtensions


ysmoradi:
I think we should check that member's type is dto or not, if not, this ignore is not required.

PR #61, opened 2017-06-20
Support regex based redirect uri + post redirect uri validation in sso + regex sample

1saeedsalehi:


ysmoradi:
Thanks @1saeedsalehi I really appreciated that (-:

PR #71, opened 2017-07-31
added ignoreChanges to delta Extensions class

Alikhll:


ysmoradi:
Thanks @Alikhll (:
It needs some enhancements, I'm gonna merge it after I added those enhancements.

PR #98, opened 2018-05-24
updated getting started contents

msynk:


ysmoradi:
You can also add VS 2017.7+ for asp.net core 2.1

PR #100, opened 2018-06-09
Depricated IAspNetCoreAppModule

behzad888:


ysmoradi:
I've merged both IAspNetCoreAppModule and IOwinAppModule into IAppModule, and it's going to be released soon.
Thanks for your attention. We're thankful about that.

ysmoradi:
It seems that we've to update our working .net core private build to the stable release of .net core 2.1 too.

PR #190, opened 2020-07-10
Update odata.md

fericode:
typo 
singalr => SignalR


fericode:
typo
singalr => SignalR

PR #201, opened 2020-09-13
Add OData article in blog

fericode:
Add an article with the title: 
**Integrate OData in .Net Core 3.1 Project With Just 20 Lines of Codes!**


docs/blog/integrate odata in .net core 3.1 project.md
```diff
@@ -0,0 +1,91 @@
+﻿# Integrate OData in .Net Core 3.1 Project With Just 20 Lines of Codes!
+
+Duo to the announcement of `Microsoft.AspNetCore.OData V7.4`, I decided to prepare an article to demonstrate how to use OData in ASP.NET Core 3.1+ projects with minimal changes in the code, so you can see how much it became easy to use in new versions. This approach is valid for projects that have endpoint routing which is the recommended approach.
+
+First with `dotnet --version` or `dotnet --info` commands or any other methods, we should be sure about installation of dot net core 3.1 sdk. Then we execute the below command:
```
ysmoradi, line 5: 3.1+

docs/blog/integrate odata in .net core 3.1 project.md
```diff
@@ -0,0 +1,91 @@
...
+
+    public string PostalCode { get; set; }
+}
+
+public enum Gender
+{
+    Man, Woman, Other
+}
+```
+For returning only woman clients: *(Filtering on Enum)*
```
ysmoradi, line 72: women customers

docs/blog/integrate odata in .net core 3.1 project.md
```diff
@@ -0,0 +1,91 @@
...
+To get customers with Ali in their names:
+```csharp
+?$filter=contains(FirstName,'Ali')
+```
+To see more querying examples in OData check out [this link](https://www.odata.org/getting-started/basic-tutorial/#queryData). 
+There are a lot of querying capabilities including Any/All in the filter query provided by OData.
+
+In addition, if we provide an IQueryable for OData instead of an array or list, the OData will execute the queries of paging, orderby, etc against the database and only the final result will return from the database, therefore it will have a huge effect on application performance.
+
+The security in this method will not be in danger, because if you put a where on the server-side query,  based on app logic, the client only can (OData) query on the limited with where applied data, so it can fetch only accessed data from the database.
```
ysmoradi, line 89: if you put a where on the server-side query or any kind of row level security,

ysmoradi (changes requested):

ysmoradi (approved):

PR #213, opened 2021-01-03
Blazor Best practices with BlazorDualMode article

fericode:
Here is the article. Please modify it anywhere you deem necessary.


fericode:
Blazor Best Practice With BlazorDualMode

ysmoradi (approved):

PR #224, opened 2021-02-06
CI action for src/Client/Web

FarhadGhE:
This YAML file will check every incoming pull request to master. If any file in `src/Client/Web/` is seen within the commit, then the build and tests of `BlazorUI` are checked.
If the flow requires a merge to develop before merging develop to master, please inform me to update the file.

### DIFF
--- a/.github/workflows/Bit.Client.Web.BlazorUI.Action.yml
+++ b/.github/workflows/Bit.Client.Web.BlazorUI.Action.yml
@@ -1,55 +1,12 @@
-name: .NET
-
+name: BlazorUI
 on:
   push:
-    branches: [ master ]
-  pull_request:
-    branches: [ master ]
-
+    paths:
+    - 'src/Client/Web/**'
 jobs:
-  check:
-    name: Check changed files
-    outputs:
-      run_job: ${{ steps.check_files.outputs.run_job }}
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout code
-        uses: actions/checkout@v2
-        with:
-          fetch-depth: 2
-
-      - name: check files
-        id: check_files
-        run: |
-          echo "=============== list changed files ==============="
-          git diff --name-only HEAD^ HEAD
-          
-          echo "========== check paths of changed files =========="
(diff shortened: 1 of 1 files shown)


FarhadGhE:
I've updated the YAML. It now uses built-in triggers of GitHub, instead of bash scripts. Also, the branch doesn't matter now.

.github/workflows/Bit.Client.Web.BlazorUI.Action.yml
```diff
@@ -0,0 +1,25 @@
...
+    name: Test successful build and Run Tests
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup .NET
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: 5.0.x
+    - name: Restore dependencies
+    #I handled execution with "find | grep | xarg", to allow later bulk operations, with reducing the regex limits
```
FarhadGhE, line 17: If simple execution is preferred, inform me to update. I wrote it for cases when all tests and all projects in the folder require build/pass check. Then added BlazorUI in the pattern to let it be specific for now.

ysmoradi: I think ```dotnet test``` step is enough because it will restore nuget packages and build project dependencies
It's also fine to hard-code project name and path

ysmoradi (commented):
Thanks @FarhadGhe I really appreciate that.
Could you please mention which issue is gonna be closed using this PR?

ysmoradi (approved):

FarhadGhE:
This pull request was created for this [issue](https://github.com/bitfoundation/bitframework/issues/221)

.github/workflows/Bit.Client.Web.BlazorUI.Action.yml
FarhadGhE: You are correct. There is still a reason to justify the separation of commands. When a PR fails the test, it's faster to monitor how this PR is failed (in which step). Still, if this isn't worth the separation, inform me to update the file.

PR #228, opened 2021-02-08
added bit-link component.

HamedMoghadasi:
1. Implemented bit-label component.
2. wrote automated tests.
3. added component to Playground project.


HamedMoghadasi:
base on **href** value, bit-link render two different element:
 1. if href **has** value, it will render **<a>** element.
 2. if href **has not** value, it will render **<button>** element.

Issue #248, opened 2021-03-12
Visibility parameter


msynk:
One of the basic and required attributes for components in a component library is Visibility.
It can be implemented like the Visibility property of UIElement in WPF: 
https://docs.microsoft.com/en-us/dotnet/api/system.windows.uielement.visibility
https://docs.microsoft.com/en-us/dotnet/api/system.windows.visibility

which is something like this in essence:
- Collapsed: Do not display the element, and do not reserve space for it in layout.
- Hidden: Do not display the element, but reserve space for the element in layout.
- Visible: Display the element.

PR #249, opened 2021-03-12
feat: add Visibility parameter (#248)

msynk:


src/Client/Web/Bit.Client.Web.BlazorUI/Components/BitComponentBase.cs
```diff
@@ -37,13 +35,29 @@ public override Task SetParametersAsync(ParameterView parameters)
...
 
                     case nameof(Class):
                         Class = (string)parameter.Value;
                         break;
+
+                    case nameof(Visibility):
+                        Visibility = (ComponentVisibility)parameter.Value;
+                        if (Visibility == ComponentVisibility.Hidden)
+                        {
+                            Style += ";visibility:hidden";
```
ysmoradi, line 54: What happens if I change Visibility multiple times?

msynk: hmmm, your right.
lemme check.
it does not affect the desired functionality but it might append that phrase multiple times.

msynk: I checked it and unfortunately it does append that phrase multiple times.
I'm working on a fix now.

ysmoradi: Maybe we need class mapper / style mapper which returns desired Style/Class as string based on added / removed classes styles.

msynk: You're right, and I suggest  to discussed it in another issue.
I pushed a fix for this issue for now.

ysmoradi (approved):

Issue #283, opened 2021-04-20
Remove extra code from BitMessageBar


ysmoradi:
StyleClass property is unused

msynk:
closed by (PR #285)

PR #285, opened 2021-04-20
Removed StyleClass property from BitMessageBar

hamedhajiloo:
Based on this #283 issue, I want to remove StyleClass property from BitMessageBar


msynk:
The issue which this PR is closing (#283) is not mentioned in the title, so the issue is not closed automatically.

Issue #310, opened 2021-04-27
missing root element for bit link component 


khashayar-pakkhesal:
bit link component needs to have a root element instead of being implemented with two separate tags and classes

PR #329, opened 2021-04-29
 Add bitlink root element (#310)

khashayar-pakkhesal:


src/Client/Web/Bit.Client.Web.BlazorUI/Components/Links/BitLink.razor
```diff
@@ -1,24 +1,23 @@
...
-    <button @ref="RootElement"
-            style="@StyleBuilder.Value"
-            class="@ClassBuilder.Value"
-            type="button"
-            @onclick="HandleClick"
-            @onclick:preventDefault="true">
-        @ChildContent
-    </button>
-}
+<span @ref="RootElement"
```
msynk, line 4: Is "Span" an appropriate root element in this situation?

khashayar-pakkhesal: I chose span since in this case root element has to have an inline display and both< button> and <a> can be a child of span

msynk (approved):

PR #333, opened 2021-04-29
fix async JSRuntime.SetProperty issue in BitCheckbox.razor.cs (#332)

msynk:


ysmoradi (approved):
It won't help much anyway (":

Issue #307, opened 2021-04-27
Add SetParametersAsync code generate for CascadingParameter properties 


mirza-developer:
Add SetParametersAsync code generating for CascadingParameter properties that can work with custom components or classes that want to use them as value of CascadingValue component
Code generating module works well with ParameterAttribute properties now

PR #354, opened 2021-05-05
Add SetParametersAsync code generate #307

Cyrus-Sushiant:


msynk (changes requested):
there are warnings in the build that should be addressed before the merge.
and please do not apply changes in the source code that are not discussed in the issue (#307 here).

src/Tooling/Bit.Tooling.SourceGenerators/Client/BlazorSetParametersSourceGenerator.cs
```diff
@@ -30,15 +30,14 @@ private string GeneratePartialClassToOverrideSetParameters(INamedTypeSymbol clas
...
-using Microsoft.AspNetCore.Components;
+            StringBuilder source = new StringBuilder($@"using Microsoft.AspNetCore.Components;
 using Microsoft.AspNetCore.Components.Web;
 using System.Threading.Tasks;
 using System;
 
 namespace {namespaceName}
 {{
-    public partial class {classSymbol.Name}  
+    public{(classSymbol.IsAbstract ? " abstract" : null)} partial class {classSymbol.Name}
```
ysmoradi, line 40: It's not required!

ysmoradi (changes requested):

src/Tooling/Bit.Tooling.SourceGenerators/Client/BlazorSetParametersSourceGenerator.cs
```diff
@@ -14,9 +14,9 @@ public void Execute(GeneratorExecutionContext context)
             if (context.SyntaxContextReceiver is not BlazorParameterPropertySyntaxReceiver receiver)
                 return;
 
-            foreach (IGrouping<INamedTypeSymbol, IPropertySymbol> group in receiver.Properties.GroupBy(f => f.ContainingType))
+            foreach (IGrouping<ISymbol, IPropertySymbol> group in receiver.Properties.GroupBy(f => f.ContainingType, SymbolEqualityComparer.Default))
             {
-                string classSource = GeneratePartialClassToOverrideSetParameters(group.Key, group.ToList());
+                string classSource = GeneratePartialClassToOverrideSetParameters(group.Key as INamedTypeSymbol, group.ToList());
```
ysmoradi, line 19: Please use cast instead of as
We use as in rare cases in which there is uncertainty about the object's real type.

ysmoradi (changes requested):

src/Tooling/Bit.Tooling.SourceGenerators/Client/BlazorSetParametersSourceGenerator.cs
msynk: +1

msynk (approved):

ysmoradi (approved):

PR #372, opened 2021-05-08
Add Icon Button Component With Tooltip (#371)

SomayeEbrahimi:


src/Client/Web/Bit.Client.Web.BlazorUI/Components/Buttons/BitIconButton.razor.cs
```diff
@@ -0,0 +1,33 @@
...
+        [Parameter] public string ToolTip { get; set; }
+        [Parameter] public EventCallback<MouseEventArgs> OnClick { get; set; }
+
+        protected override string RootElementClass => "bit-ico-btn";
+
+        protected override void RegisterComponentClasses()
+        {
+            ClassBuilder.Register(() => IsEnabled is false
+                                        ? string.Empty
+                                        : $"{RootElementClass}-default-{VisualClassRegistrar()}");
```
msynk, line 22: Can't we use the enabled CSS class itself here?

PR #373, opened 2021-05-08
Add Action Button Component (#366)

SomayeEbrahimi:


src/Client/Web/Bit.Client.Web.BlazorUI/Components/Buttons/BitActionButton.razor.cs
```diff
@@ -0,0 +1,33 @@
...
+        [Parameter] public RenderFragment ChildContent { get; set; }
+        [Parameter] public EventCallback<MouseEventArgs> OnClick { get; set; }
+
+        protected override string RootElementClass => "bit-act-btn";
+
+        protected override void RegisterComponentClasses()
+        {
+            ClassBuilder.Register(() => IsEnabled is false
+                                        ? string.Empty
+                                        : $"{RootElementClass}-default-{VisualClassRegistrar()}");
```
msynk, line 22: why introducing a new CSS class? can't we just use the enabled CSS class?

PR #375, opened 2021-05-09
Add Button-like Anchor (#374)

SomayeEbrahimi:


msynk (commented):
The "DefaultButton" in Fluent UI is "BitButton" in bit framework. there is no need to introduce a new component for that matter.

src/Client/Web/Bit.Client.Web.BlazorUI.Tests/Buttons/BitDefaultButtonTest.razor
```diff
@@ -0,0 +1,24 @@
+﻿<CascadingValue Value="Visual">
```
msynk, line 1: this file is redundant

src/Client/Web/Bit.Client.Web.BlazorUI.Tests/Buttons/BitDefaultButtonTests.razor.cs
```diff
@@ -0,0 +1,92 @@
+﻿using System.Threading.Tasks;
```
msynk, line 1: this file is redundant

msynk (changes requested):

Issue #298, opened 2021-04-25
Message bar styles


khashayar-pakkhesal:
Message bar examples needs to be implemented just like fluent 
and has no icons

PR #378, opened 2021-05-10
Fix message bar icons (#298)

khashayar-pakkhesal:


msynk (changes requested):

src/Client/Web/Bit.Client.Web.BlazorUI/Components/Notifications/BitMessageBar.razor
```diff
@@ -4,7 +4,25 @@
 <div @ref="RootElement"
      style="@StyleBuilder.Value"
      class="@ClassBuilder.Value">
-    <span class="flexContainer">
+    <span>
+        <i class="bit-icon bit-icon--@MessageBarIconName"></i>
+    </span>
+    <span class="text-container @( (Truncated && !ElementTruncateState) || IsMultiline ? "not-truncate" : "")">
```
msynk, line 10: there is no need to have the "text-container" class in here.

src/Client/Web/Bit.Client.Web.BlazorUI/Styles/Fluent/Notifications/bit-message-bar.fluent.scss
```diff
@@ -11,34 +11,88 @@
...
+        -webkit-box-orient: vertical;
+        overflow: hidden;
+        -webkit-line-clamp: 1;
+        margin: 8px 0px 8px 12px;
+        padding: 0;
+        text-align: left;
+    }
+
+    .not-truncate {
+        -webkit-line-clamp: unset;
```
msynk, line 28: do not use "-webkit-line-clamp" to truncate the text. see the Fluent UI sample to get an idea about the correct way of implementing it using overflow and text-overflow properties.

msynk (changes requested):

src/Client/Web/Bit.Client.Web.BlazorUI/Components/Notifications/BitMessageBar.razor
khashayar-pakkhesal: in here we have 3 spans one for left side icon one for right side icons and the other one which is in thee middle is for text I thought it's better that we assign a class to the one containing our text

msynk: yes, you are right, but as we spoke about these issues before, because the structure of the component is clear and constant for us (the developers) we can use the ":nth-child()" selector here.

khashayar-pakkhesal: Fixed

src/Client/Web/Bit.Client.Web.BlazorUI/Styles/Fluent/Notifications/bit-message-bar.fluent.scss
khashayar-pakkhesal: Fixed

msynk (approved):

Issue #400, opened 2021-05-16
Failed Unit Test of BitButton


SomayeEbrahimi:
After new changes in button component, its test did not pass and show this exception : MissingEventHandlerException

PR #401, opened 2021-05-16
Fix issue in bit button test (#400)

SomayeEbrahimi:


src/Client/Web/Bit.Client.Web.BlazorUI.Tests/Buttons/BitButtonLinkTest.razor
```diff
@@ -0,0 +1,24 @@
+﻿<CascadingValue Value="Visual">
```
msynk, line 1: This approach to resolving the Test issues is not accepted.
please consider the correct solution (resolving the ambiguity of the "string" parameter using @ character) as we talked about in our meeting.

msynk (changes requested):

src/Client/Web/Bit.Client.Web.BlazorUI.Tests/Buttons/BitButtonTest.razor
```diff
@@ -2,12 +2,12 @@
...
-               AriaHidden="AriaHidden"
-               Href="Href"
-               Title="Title"
-               Target="Target">
+               AriaDescription="@AriaDescription"
+               AriaLabel="@AriaLabel"
+               AriaHidden="@AriaHidden"
+               Href="@Href"
+               Title="@Title"
+               Target=@Target>
```
msynk, line 10: Is there any reason for only the "Target" parameter's value has no double quotation mark?

src/Client/Web/Bit.Client.Web.BlazorUI.Tests/Buttons/BitButtonTest.razor
```diff
@@ -2,12 +2,12 @@
...
                ButtonStyle="ButtonStyle"
-               AriaDescription="AriaDescription"
-               AriaLabel="AriaLabel"
-               AriaHidden="AriaHidden"
-               Href="Href"
-               Title="Title"
-               Target="Target">
+               AriaDescription="@AriaDescription"
+               AriaLabel="@AriaLabel"
+               AriaHidden="@AriaHidden"
```
msynk, line 7: the "AriaHidden" parameter does not require the "@" in its value, because its type is not string

msynk (changes requested):

ysmoradi:
Thanks so much @SomayeEbrahimi 
Working on tests is really appreciated!

PR #410, opened 2021-05-19
Add AllowedDisabledFocus attribute on BitButton component (#379)

SomayeEbrahimi:


src/Client/Web/Playground/Bit.Client.Web.BlazorUI.Playground/Web/Pages/Component.razor
```diff
@@ -35,7 +35,7 @@
     <BitButton IsEnabled="false">
         Disabled
     </BitButton>
-    <BitButton IsEnabled="false">
+    <BitButton IsEnabled="false" AllowDisabledFocus="false">
```
msynk, line 38: Please provide more samples of this parameter

src/Client/Web/Bit.Client.Web.BlazorUI.Tests/Buttons/BitButtonTests.cs
```diff
@@ -46,6 +46,28 @@ public Task BitButtonTest(Visual visual, bool isEnabled, ButtonStyle style)
...
+            var com = RenderComponent<BitButtonTest>(parameters =>
+            {
+                parameters.Add(p => p.IsEnabled, isEnabled);
+                parameters.Add(p => p.ButtonStyle, style);
+                parameters.Add(p => p.AllowDisabledFocus, allowDisabledFocus);
+            });
+
+            var bitButton = com.Find(".bit-btn");
+
+            Assert.AreEqual(bitButton.HasAttribute("tabindex"), expectedResult);
```
msynk, line 66: please add a check to the value of the "tabindex" attribute too.
Its value should be equal to "-1".

msynk (changes requested):

msynk (changes requested):

src/Client/Web/Bit.Client.Web.BlazorUI.Tests/Buttons/BitButtonTests.cs
```diff
@@ -63,7 +63,8 @@ public Task BitButtonDisabledFocusTest(bool isEnabled, ButtonStyle style, bool a
 
             var bitButton = com.Find(".bit-btn");
 
-            Assert.AreEqual(bitButton.HasAttribute("tabindex") ? bitButton.GetAttribute("tabindex").Equals("-1") : expectedResult, expectedResult);
+            if (bitButton.HasAttribute("tabindex"))
+                Assert.AreEqual(bitButton.GetAttribute("tabindex").Equals("-1"), expectedResult);
```
msynk, line 67: this solution is not correct either. there should be 2 assertions in these tests:
The first assertion should be about checking the existence of a "tabindex" attribute (as before).
The second assertion should be about the value of "tabindex" if the value of expectedResult is true.

msynk (changes requested):

msynk (approved):

PR #441, opened 2021-06-07
Resolve 2-way data binding issue in BitCheckBox component when the parameter is not provided (#440)

### LINKED ISSUE #440: Two way data Binding issue in BitCheckBox
consumers can work with value in two ways 
1. consumer can get the value of BitCheckBox every time it gets changed using OnChange and in this scenario we provide the component with full functionality
2. consumer can control the value of BitCheckBox using both IsChecked and IsCheckedChanged and it is necessary to provide them both in order to use the full functionality of the component if the consumer provides us with IsChecked we cannot change the value of the component from inside 

currently, OnChange doesn't work and we have to change our approach to implementing two-way data binding

khashayar-pakkhesal:
this closes #440


src/Client/Web/Bit.Client.Web.BlazorUI/Components/Checkboxes/BitCheckbox.razor.cs
```diff
@@ -61,6 +61,8 @@ public bool IsIndeterminate
 
         [Parameter] public EventCallback<bool> OnChange { get; set; }
 
+        private bool hasSetIsChecked { get; set; }
```
msynk, line 64: It is not a good name for this property, I think it should be named something like "IsCheckedHasValue" or "IsCheckHasBeenSet"

msynk (changes requested):

src/Client/Web/Bit.Client.Web.BlazorUI/Components/Checkboxes/BitCheckbox.razor.cs
msynk: And change this property into a private field

src/Client/Web/Bit.Client.Web.BlazorUI/Components/Checkboxes/BitCheckbox.razor.cs
```diff
@@ -108,5 +110,38 @@ protected override async Task OnAfterRenderAsync(bool firstRender)
 
             await base.OnAfterRenderAsync(firstRender);
         }
+
+        public override Task SetParametersAsync(ParameterView parameters)
+        {
+            foreach (ParameterValue parameter in parameters)
+            {
+                switch (parameter.Name)
```
msynk, line 118: The private field holding the state of value of 2-way bound parameter should be reset at the start of this method

src/Client/Web/Bit.Client.Web.BlazorUI/Components/Checkboxes/BitCheckbox.razor.cs
khashayar-pakkhesal: Changed hasSetIsChecked  to isCheckedHasValue

src/Client/Web/Bit.Client.Web.BlazorUI/Components/Checkboxes/BitCheckbox.razor.cs
khashayar-pakkhesal: Done

src/Client/Web/Bit.Client.Web.BlazorUI/Components/Checkboxes/BitCheckbox.razor.cs
msynk: after some investigation, I think the "IsCheckedHasBeenSet" is a better name. please apply this particular pattern for naming this specific field, which will be applied in the source generator in near future.

khashayar-pakkhesal: Rename hasSetIsChecked to IsCheckedHasBeenSet

msynk (approved):

Issue #452, opened 2021-06-10
Add BitLabel component demo page


SomayeEbrahimi:
This issue is based on issue #278

PR #483, opened 2021-06-17
Add BitLabel component demo page (#452)

SomayeEbrahimi:


src/Client/Web/Playground/Bit.Client.Web.BlazorUI.Playground/Web/Pages/Components/NavMenu.razor.cs
```diff
@@ -16,7 +16,7 @@ public partial class NavMenu
                      new NavLink { Name= "Button", Key = "Button", Url = "/components/buttons", TargetType = NavLinkTargetType.Self },
                      new NavLink { Name= "Checkbox", Key = "Checkbox", Url = "#" },
                      new NavLink { Name= "Choice", Key = "Choice", Url = "#" },
-                     new NavLink { Name= "Label", Key = "Label", Url = "#" },
+                     new NavLink { Name= "Label", Key = "Label",  Url = "/components/labels", Target = "_Self" },
```
msynk, line 19: the Target parameter should not be set here, because this value causes a hard refresh of the application

msynk (changes requested):

PR #499, opened 2021-06-22
Fix Autogenerating SetParametersAsync method for Generic Component #408

### LINKED ISSUE #408: Autogenerating SetParametersAsync method for Generic Component
The current source generator solution does not account for the generic parameter in a generic component.
for example, consider a generic component class with the following structure:

```cs
public partial class BitBasicList<TItem>
{
  ...
}
```
the generated class is as follows:
```cs
public partial class BitBasicList
{
  public override Task SetParametersAsync(ParameterView parameters)
  {
    ...
  }
}
```
which is not correct.

Cyrus-Sushiant:
This closes #408


msynk:
the branch name is not totally compatible with our branch naming conventions :)
Please review other PRs in order to find out about your mistakes here.

msynk (approved):

PR #529, opened 2021-06-29
Make the whole area of parent BitNavLink element clickable (#495)

### LINKED ISSUE #495: The whole area of parent NavLink element should be clickable
The whole area of parent NavLink element should be clickable

as showing in the following image, only the chevron icon is clickable in order to open the node:

image

it is required that the whole area of this parent element to be clickable and active, both for opening the node and also for navigation.

zoha-shobbar:
This closes #495


src/Client/Web/Bit.Client.Web.BlazorUI/Components/Navs/BitNav.razor
```diff
@@ -49,12 +49,11 @@
 
         childTemplate = (navLink) =>
         (
-    @<li role="listitem" @key="navLink.Key">
+    @<li role="listitem" @key="navLink.Key" @onclick="(() => Toggle(navLink))">
```
msynk, line 52: this is not the correct way to fix this issue, it is making another issue that is related to bubbling the click event to this parent element from all of the child elements.

src/Client/Web/Bit.Client.Web.BlazorUI/Styles/Fluent/Navs/bit-nav.fluent.scss
```diff
@@ -19,7 +19,6 @@
     user-select: none;
 
     @if $hasUrl != true {
-        pointer-events: none;
```
msynk, line 22: why did you remove this line?

msynk (changes requested):

msynk:
I'm closing this PR because of the unrelated commits in it

PR #666, opened 2021-08-08
Modify BitIconButton component features (#663)

### LINKED ISSUE #663: Modify BitIconButton features
We should add "type" attribute to button tag which is equal to "button":

freesnippingtool com_capture_20210808172338
---

Hover mode for BitIconButton is not implemented:

freesnippingtool com_capture_20210808172653
----

If an "href" attribute is given by the user, we should render an "a" tag instead of a "button" and some test methods are required for this feature:

freesnippingtool com_capture_20210808172837
---

Some css attributes are missed:

freesnippingtool com_capture_20210808173946
---

It is better to change the name of "ToolTip" parameter of BitIconButton to "Title":

freesnippingtool com_capture_20210808172917

maryamhdr:
This closes #663

### DIFF
--- a/src/Client/Web/Bit.Client.Web.BlazorUI.Tests/Buttons/BitIconButtonTests.cs
+++ b/src/Client/Web/Bit.Client.Web.BlazorUI.Tests/Buttons/BitIconButtonTests.cs
@@ -23,14 +23,14 @@ namespace Bit.Client.Web.BlazorUI.Tests.Buttons
                DataRow(Visual.Material, true, "Emoji2", "I'm Happy", true),
                DataRow(Visual.Material, false, "Emoji2", "I'm Happy", false),
            ]
-        public void BitIconButtonTest(Visual visual, bool isEnabled, string iconName, string toolTip, bool expectedResult)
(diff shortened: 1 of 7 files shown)


src/Client/Web/Bit.Client.Web.BlazorUI.Tests/Buttons/BitIconButtonTests.cs
```diff
@@ -23,14 +23,14 @@ public class BitIconButtonTests : BunitTestContext
                DataRow(Visual.Material, true, "Emoji2", "I'm Happy", true),
                DataRow(Visual.Material, false, "Emoji2", "I'm Happy", false),
            ]
-        public void BitIconButtonTest(Visual visual, bool isEnabled, string iconName, string toolTip, bool expectedResult)
+        public void BitIconButtonTest(Visual visual, bool isEnabled, string iconName, string title, bool expectedResult)
```
msynk, line 26: please remove the "expectedResult" argument and calculate its desired value in the test method body based on the logic we expect to be applied.

msynk (changes requested):

msynk (approved):

PR #673, opened 2021-08-09
Modify BitToggleButton component features (#670) 

### LINKED ISSUE #670: Modify BitToggleButton features
We should add "type" attribute to button tag which is equal to "button":
freesnippingtool com_capture_20210809232517
---

If an "href" attribute is given by the user, we should render an "a" tag instead of a "button" and some test methods are required for this feature:
freesnippingtool com_capture_20210809232604
---

Some css attributes are missed:
freesnippingtool com_capture_20210809233205
---

The "Title" parameter should be added to the component:
freesnippingtool com_capture_20210809232944
---

Disabled mode in Standard Button is different from Primary Button:
freesnippingtool com_capture_20210809232845

freesnippingtool com_capture_20210809232757
---

If no value is provided for IconName, the "i" tag should not be rendered. The same issue happens while the "Label" parameter is not provided:
freesnippingtool com_capture_20210809234038
---

The style of the button is incorrect while IsChecked is equal to true:
freesnippingtool com_capture_20210809234117
freesnippingtool com_capture_20210809234153

maryamhdr:
This closes #670


msynk:
I'm closing this PR because of unrelated commits in it

PR #752, opened 2021-09-03
Fix close problem of multi-select on item click

### LINKED ISSUE #528: Multi-select dropdown should not be closed when we click on items
base on [fluent UI](https://developer.microsoft.com/en-us/fluentui#/controls/web/dropdown), When we select an item in the multi-select dropdown, it should stay open.

RezaKargar:
This fixes #528, the multi-select dropdown must
not be closed on its items click


msynk:
since the issue is resolved in another PR (#763) with a more robust solution, I'm closing this PR.

PR #803, opened 2021-09-15
Fix url comparison in BitNav component (#785)

### LINKED ISSUE #785: BitNav Selected Url Comparison
We need to find a way to completely compare the current page address with BitNav button URL to find the selected one,
 our current code does not meet the expectations on some Urls

RezaKargar:
This commit closes #785,
Now selected link will be highlighted correctly and also
selectedkey will perform right


msynk:
since the related issue is addressed in another PR, I'm going to close this PR.

PR #938, opened 2021-10-25
Fix BitDropDown wrapper position on scroll screen (#430)

### LINKED ISSUE #430: BitDropDown direction of opening the callout based on its position
BitDropDown should automatically check its position and the available space on the screen to calculate the correct direction to open in order to show a full view of the callout element which is not applying for now. an example of this issue is showing in the following image:

image

mirza-developer:
This closes #430


msynk:
since there was no activity for this PR in a long time, I'm closing it.

PR #1017, opened 2021-11-07
Add header functionality in playground project (#999)

### LINKED ISSUE #999: Add header functionality in playground project
Based on the new design there are some links for navigating to different pages in the header of the playground project, and in each page, the related link page should be active. so it should be added as a component to use on all pages.

SomayeEbrahimi:
This close #999

### DIFF
--- a/src/Client/Web/Playground/Bit.Client.Web.BlazorUI.Playground/Web/Pages/Components/Header.razor
+++ b/src/Client/Web/Playground/Bit.Client.Web.BlazorUI.Playground/Web/Pages/Components/Header.razor
@@ -9,13 +9,13 @@
     </div>
     <ul class="header-menu">
         <li>
-            <BitLink Href="\">Home</BitLink>
+            <BitLink Class="@homeLinkClass" Href="\">Home</BitLink>
         </li>
         <li>
-            <BitLink Class="active" Href="\components\overview">Demo</BitLink>
+            <BitLink Class="@demoLinkClass" Href="\components\overview">Demo</BitLink>
         </li>
         <li>
-            <BitLink Href="\getstarted">Get Started</BitLink>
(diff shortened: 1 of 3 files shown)


msynk (changes requested):
as we discussed earlier, plz refactor the current implementations.

src/Client/Web/Playground/Bit.Client.Web.BlazorUI.Playground/Web/Pages/Components/Header.razor.cs
```diff
@@ -1,6 +1,21 @@
-﻿namespace Bit.Client.Web.BlazorUI.Playground.Web.Pages.Components
+﻿using Microsoft.AspNetCore.Components;
+
+namespace Bit.Client.Web.BlazorUI.Playground.Web.Pages.Components
 {
     public partial class Header
     {
+        private readonly string demoPath = "components";
+        private readonly string getStartedPath = "get-started";
```
msynk, line 8: there is no need to define these variables, plz move the strings into the razor file.

msynk (changes requested):

src/Client/Web/Playground/Bit.Client.Web.BlazorUI.Playground/Web/Pages/Components/Header.razor
```diff
@@ -12,10 +12,10 @@
             <BitLink Href="\">Home</BitLink>
         </li>
         <li>
-            <BitLink Class="active" Href="\components\overview">Demo</BitLink>
+            <BitLink Class="@((currentUri.StartsWith("components")) ? "active" : string.Empty)" Href="\components\overview">Demo</BitLink>
```
msynk, line 15: doesn't all page URLs start with "components"?

msynk (changes requested):

msynk (approved):

PR #1100, opened 2021-11-20
refactor file structure #1099

### LINKED ISSUE #1099: Improve file structure of the components project based on the Fluent UI react library
The current file structure of the components project is not totally compatible with the Fluent UI react library.
it is better to be refactored in order to make the structure as much as possible compatible with that library which is available on GitHub:
https://github.com/microsoft/fluentui/tree/master/packages/react/src/components

msynk:
this closes #1099


ysmoradi:
Wow!

maryamhdr (approved):

PR #1130, opened 2021-12-05
fix: update state text in BitToggle's ui based on changes of IsChecked parameter

### LINKED ISSUE #1129: OnText/OffText are not getting updated after initialization in BitToggle component
after initialization in the BitToggle, the values set for the `OnText` and `OffText` parameters are not getting updated in the UI with the user interaction and changing the IsChecked parameter.

msynk:
this closes #1129

### DIFF
--- a/src/Client/Web/Bit.Client.Web.BlazorUI/Components/Toggle/BitToggle.razor.cs
+++ b/src/Client/Web/Bit.Client.Web.BlazorUI/Components/Toggle/BitToggle.razor.cs
@@ -62,6 +62,7 @@ namespace Bit.Client.Web.BlazorUI
             {
                 if (value == isChecked) return;
                 isChecked = value;
+                SetTexts();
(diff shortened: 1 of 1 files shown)


src/Client/Web/Bit.Client.Web.BlazorUI/Components/Toggle/BitToggle.razor.cs
```diff
@@ -113,10 +121,8 @@ protected override async Task OnInitializedAsync()
                 if (StateText.HasValue())
                 {
                     LabelledById = LabelledById.HasValue() ? LabelId + " " + StateTextId : StateTextId;
```
Cyrus-Sushiant, line 123: Replace with string interpolation.
`$"{LabelId} {StateTextId}"`

Cyrus-Sushiant (changes requested):

PR #1217, opened 2021-12-20
fix BitDropDown component scroll issue (#1146)

### LINKED ISSUE #1146: Fix BitDropDown component's callout scroll issue
The issue is that the drop-down callout should not be closed while scrolling.

maryamhdr:
this closes #1146


src/Client/Web/Bit.Client.Web.BlazorUI/Scripts/bit.blazorui.ts
```diff
@@ -51,6 +51,8 @@ class Bit {
     }
 }
 
-window.addEventListener('scroll', () => {
+window.addEventListener('scroll', (e: any) => {
+    if (e.target.id && e.target.id.toString().toLowerCase().indexOf("dropdown") > -1) return;
```
msynk, line 55: it's better to use a more reliable method in this condition

msynk (changes requested):

msynk (approved):

PR #1223, opened 2021-12-21
Modify BitDatePicker RTL direction 

### LINKED ISSUE #1220: RTL support in BitDatePicker is not complete
The Jalali date picker should get the RTL direction so the styles of its elements should be modified to catch this purpose.

maryamhdr:
this closes #1220


src/Client/Web/Bit.Client.Web.BlazorUI/Components/DatePicker/BitDatePicker.razor.cs
```diff
@@ -156,6 +156,9 @@ protected override void RegisterComponentClasses()
         {
             ClassBuilder.Register(() => IsEnabled is false
                 ? $"{RootElementClass}-disabled-{VisualClassRegistrar()}" : string.Empty);
+
+            ClassBuilder.Register(() => Culture.TextInfo.IsRightToLeft
```
msynk, line 160: since the `Culture` parameter has direct effect on the `ClassBuilder`, it should reset the `ClassBuilder` after setting new values.

msynk (changes requested):

PR #1261, opened 2022-01-03
Added _Layout.cshtml to bit components playground

### LINKED ISSUE #1260: Use _Layout.cshtml in Bit Components Playground

ysmoradi:
This closes #1260


src/Client/Web/Playground/Bit.Client.Web.BlazorUI.Playground/Web/App.razor
```diff
@@ -1,4 +1,5 @@
-﻿<CascadingValue Value="Visual.Fluent" IsFixed="true">
+﻿<PageTitle>Bit's Native Blazor UI Components</PageTitle>
```
ysmoradi, line 1: Just to let you know that from now, we can use <PageTitle> in our razor pages so we can use the customized titles/meta tags for each page. @msynk @maryamhdr

PR #1337, opened 2022-01-18
BitChoiceGroup disabled mode issue (#1199)

### LINKED ISSUE #1199: BitChoiceGroup disabled mode issue
Choice options in a disabled choice group should not get a selected state.

freesnippingtool com_capture_20211215124123

saeedhalab:
This closes #1199


maryamhdr (changes requested):
This issue is that the BitChoiceGroup component in disabled mood must not have a selected item even if the SelectedKey has been set by the user. 
In another word, we should check if the component is disabled, the selected key be empty even if the user provides some SelectedKey as the component parameter.

saeedhalab:
in Fluent site  When set  ChoiceGroup Is disabled mood and user set DefaultSelectedKey we have selected item even but selected item color is different

msynk (approved):

PR #1341, opened 2022-01-19
Implement sign-in page UI(#1266)

### LINKED ISSUE #1266: Implement sign-in page UI
Implement the sign-in page UI.

saeedhalab:
This closes #1266

### DIFF
--- a/src/Tooling/Bit.Tooling.Templates/TodoTemplate/Web/Components/Header.razor
+++ b/src/Tooling/Bit.Tooling.Templates/TodoTemplate/Web/Components/Header.razor
@@ -1,7 +1,7 @@
-﻿<header class="header @(true ? "header--authenticated" : null)">
+﻿<header class="header @(false ? "header--authenticated" : null)">
     <div class="header-top-row">
         <div class="header-contact">
-            @if (true)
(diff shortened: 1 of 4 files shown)


maryamhdr (approved):

src/Tooling/Bit.Tooling.Templates/TodoTemplate/Web/Components/Header.razor
```diff
@@ -1,18 +1,18 @@
-﻿<header class="header @(true ? "header--authenticated" : null)">
+﻿<header class="header @(false ? "header--authenticated" : null)">
     <div class="header-top-row">
         <div class="header-contact">
-            @if (true)
+            @if (false)
```
ysmoradi, line 4: ? @maryamhdr @saeedhalab

maryamhdr: The purpose of using this `if` condition is to handle the header state if the user is authenticated or not.
We will have a service in the future in which we will handle the user authentication state. Till then, to be able to implement the UI for every state, we have used the very condition.

PR #1363, opened 2022-01-27
Implement Model to-do (#1360)

### LINKED ISSUE #1360: Implement Model to-do
in this issue, we should create model for to-do and add to-do table  to SQL server Db

saeedhalab:
This closes #1360


src/Tooling/Bit.Tooling.Templates/TodoTemplate/Api/Data/Models/Todo/Todo.cs
```diff
@@ -0,0 +1,12 @@
+﻿using TodoTemplate.Shared.Enums;
+
+namespace TodoTemplate.Api.Data.Models.Todo
+{
+    public class Todo
```
msynk, line 5: it's better to not use these general names for our models.
a better name would be `TodoItem`

src/Tooling/Bit.Tooling.Templates/TodoTemplate/Api/Data/Models/Todo/Todo.cs
```diff
@@ -0,0 +1,12 @@
...
+    {
+        public int Id { get; set; }
+        public string Title { get; set; }
+        public DateTime Date { get; set; }
+        public TodoState State { get; set; }
```
msynk, line 10: since there is no other state for this property, it's better to use a boolean one named `IsDone`

msynk (changes requested):

msynk (approved):

PR #1380, opened 2022-02-03
implement pages authorization based on token (#1371)

### LINKED ISSUE #1371: implement pages authorization based on token
We must apply the access level to all pages so that if a user does not have access to a page they can not see it. Also, if the user does not have access, we must move it to the sign-in page

sajadkardel:
this closes #1371


src/Tooling/Bit.Tooling.Templates/TodoTemplate/Web/App.razor
```diff
@@ -1,11 +1,18 @@
...
-		</LayoutView>
-	</NotFound>
-</Router>
+﻿<CascadingAuthenticationState>
+    <PageTitle>To Do app!</PageTitle>
+    <Router AppAssembly="@GetType().Assembly">
+        <Found Context="routeData">
+            <AuthorizeRouteView RouteData="@routeData" DefaultLayout="@typeof(MainLayout)">
+                <NotAuthorized>
+                    <RedirectToSignIn />
```
ysmoradi, line 7: Why not using sign-in component directly here?

ysmoradi (changes requested):

src/Tooling/Bit.Tooling.Templates/TodoTemplate/Web/App.razor
msynk: It's a page, not a simple component, using pages as components is a bad practice.

msynk (approved):

src/Tooling/Bit.Tooling.Templates/TodoTemplate/Web/App.razor
msynk: but we need to handle the `redirectUrl` in this solution which will be handled in a separate issue.

msynk (approved):

PR #1402, opened 2022-02-08
enable the user to edit profile (#1288)

### LINKED ISSUE #1288: Enable the user to edit his/her profile
The user can edit his/her info and save them on the server.
Also, we should display the success or failure notifications.

sajadkardel:
this closes #1288

### DIFF
--- a/src/Tooling/Bit.Tooling.Templates/TodoTemplate/Api/Controllers/AttachmentController.cs
+++ b/src/Tooling/Bit.Tooling.Templates/TodoTemplate/Api/Controllers/AttachmentController.cs
@@ -1,0 +1,73 @@
+﻿using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Options;
+using TodoTemplate.Api.Contracts;
+
+namespace TodoTemplate.Api.Controllers
+{
+    [Route("api/[controller]")]
+    [ApiController]
+    [Authorize]
+    public class AttachmentController : ControllerBase
+    {
+        private readonly AppSettings _appSettings;
+
+        public AttachmentController(IOptionsSnapshot<AppSettings> setting)
+        {
+            _appSettings = setting.Value;
+        }
+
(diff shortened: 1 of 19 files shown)


src/Tooling/Bit.Tooling.Templates/TodoTemplate/Api/Controllers/AttachmentController.cs
```diff
@@ -0,0 +1,73 @@
...
+            if (!System.IO.File.Exists(path)) return NotFound();
+
+            System.IO.File.Delete(path);
+            return Ok();
+        }
+
+        [HttpGet("[action]/{userName}"), AllowAnonymous]
+        public ActionResult GetProfilePhoto(string userName)
+        {
+            if (!Directory.Exists(_appSettings.UserProfilePhotoPath)) return Ok();
```
ysmoradi, line 61: Please return not found

src/Tooling/Bit.Tooling.Templates/TodoTemplate/Api/Controllers/AttachmentController.cs
```diff
@@ -0,0 +1,73 @@
...
+
+            var filePath = Directory.GetFiles(_appSettings.UserProfilePhotoPath)
+                .FirstOrDefault(s => s.Contains(userName));
+
+            if (filePath is null) return Ok();
```
ysmoradi, line 66: Please return not found

ysmoradi (changes requested):

src/Tooling/Bit.Tooling.Templates/TodoTemplate/Api/Controllers/AttachmentController.cs
sajadkardel: We may have scenarios where the user does not have a profile picture. In this case, if we return the Not Found, the program will get a server error. For this reason, this action returns an OK or return file bytes.
Do you have a new idea for this?

PR #1416, opened 2022-02-10
Fix selected key issue on BitChoiceGroup #1392

### LINKED ISSUE #1392: SelectedKey is not working on the initialize of the BitChoiceGroup
when we give a value to the selectedkey parameter, before we select an option the key in the selected key variable is not selected in the first load.

zoha-shobbar:
this closes #1392


msynk:
it's not a correct fix.
Since we are going to redesign this component very soon, I'm closing this PR and the corresponding issue.

PR #1440, opened 2022-02-14
fix not found photo error on edit profile page (#1435)

### LINKED ISSUE #1435: Fix not found photo error on EditProfile page
We may have a user who does not have an image, in which case we should prevent the error and display the first two letters of the username instead of the image.

sajadkardel:
this closes #1435


ysmoradi:
Pending until this issue gets resolved https://github.com/bitfoundation/bitframework/issues/1422

PR #1453, opened 2022-02-16
Add style to the BitToggleButton with href attribute #1135

### LINKED ISSUE #1135: Incorrect style in BitToggleButton when use href parameter
In the `BitToggleButton` component when we set the `href` parameter the button takes an incorrect style like the following Image:

image

zoha-shobbar:
this closes #1135


maryamhdr (changes requested):
This issue is about the nonproper styles of the root tag of the component no matters it is an anchor tag or a button tag.
You should modify the styles of the root tag just like the Fluent UI so the issue would be solved :)

maryamhdr (approved):

msynk (approved):

PR #1499, opened 2022-02-22
fileupload failed callbacks

### LINKED ISSUE #1497: Add OnUploadFailed event in BitFileUpload component
We need an event callback which will be raised if the upload process fails. We should send the failure message back via this callback.

msynk:
this closes #1497


maryamhdr (approved):
You are the best @msynk :))))

PR #1552, opened 2022-02-28
rename TokenProvider to AuthTokenProvider (#1480)

### LINKED ISSUE #1480: Rename 'TokenProvider' to 'AuthTokenProvider'
Currently, both the interface and implementation of the access token provider are using `TokenProvider` which is very generic and kinda ambiguous. It's better to use a more specific name like `AuthTokenProvider` to remove the ambiguity.

msynk:
this closes #1480


maryamhdr (dismissed):

ysmoradi (commented):
Shall we use ClientSideAuthTokenProvider instead of ClientSideTokenProvider? @msynk

ysmoradi (approved):

PR #1571, opened 2022-03-02
Resolve issue change value with keyboard in the BitSpinButton (#1548)

### LINKED ISSUE #1548: It is not possible to change the input value with the keyboard in the BitSpinButton
The BitSpinButton has a bug. When setting the value for the Suffix parameter, it is not possible to change the input value with the keyboard.

Cyrus-Sushiant:
This closes #1548

### DIFF
--- a/src/Client/Web/Bit.Client.Web.BlazorUI/Components/SpinButton/BitSpinButton.razor.cs
+++ b/src/Client/Web/Bit.Client.Web.BlazorUI/Components/SpinButton/BitSpinButton.razor.cs
@@ -6,6 +6,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Components;
 using Microsoft.AspNetCore.Components.Web;
+using Microsoft.JSInterop;
 
 namespace Bit.Client.Web.BlazorUI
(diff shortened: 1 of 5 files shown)


src/Client/Web/Bit.Client.Web.BlazorUI/Components/SpinButton/BitSpinButton.razor.cs
```diff
@@ -17,9 +18,14 @@ public partial class BitSpinButton
...
         private double max;
+        private string? intermediateValue;
         private string InputId = $"input{Guid.NewGuid()}";
-        private string IntermediateValue = string.Empty;
         private Timer? timer;
+        private ElementReference inputRef;
+        private ElementReference buttonIncrement;
+        private ElementReference buttonDecrement;
+
+        [Inject] public IJSRuntime? JSRuntime { get; set; }
```
msynk, line 28: plz remove the nullable `?` and use `default!` as an initial value here, and also fix all `JSRuntime!` by removing the `!` character.

msynk (changes requested):

msynk (approved):

PR #1581, opened 2022-03-04
Change Program.cs of Playground.Web to use WebApplication 

### LINKED ISSUE #1579: Use `WepApplication` instead of HostBuilder for BlazorServer in the Playground project
In Bit Components Playground, web project,  we should use WebApplication in Program.cs rather than HostBuilder.

RezaKargar:
This resolves #1579 by changing **Program.cs** of **Playground.Web** to use `WebApplication` rather than `HostBuilder`


ysmoradi:
Thanks @RezaKargar 
Could you please follow the same approach that we've already taken in the Web project of Todo app template project?
https://github.com/bitfoundation/bitframework/blob/develop/src/Tooling/Bit.Tooling.Templates/TodoTemplate/Web/Program.cs

PR #1729, opened 2022-03-18
Make BitDropDown responsive in small screens (#429)

### LINKED ISSUE #429: BitDropDown should be responsive to small screens
As the default behaviour of the counterpart component in Fluent UI framework, the BitDropDown component should be opened like the following image in response to the small screens:

RezaKargar:
This closes #429

### DIFF
--- a/src/Client/Web/Bit.Client.Web.BlazorUI/Components/DropDown/BitDropDown.Fluent.scss
+++ b/src/Client/Web/Bit.Client.Web.BlazorUI/Components/DropDown/BitDropDown.Fluent.scss
@@ -290,19 +290,15 @@
 }
 
 .overflow-hidden {
-    overflow: hidden !important;
+    overflow: hidden;
 }
 
-@media only screen and (min-width: 0) and (max-width: rem(639px)) {
+@media only screen and (max-width: em(639px)) {
     .bit-drp-items-wrapper {
         box-shadow: rgb(0 0 0 / 22%) 0 25.6px 57.6px 0, rgb(0 0 0 / 18%) 0 4.8px 14.4px 0;
-        position: fixed !important;
-        top: 0 !important;
-        left: unset !important;
-        right: 0 !important;
-        bottom: unset !important;
-        height: 100% !important;
-        width: rem(272px) !important;
+        position: fixed;
+        height: 100%;
+        width: rem(272px);
         padding: em(24px) em(20px);
         z-index: 10000;
         animation-name: opacity, trans-x;
--- a/src/Client/Web/Bit.Client.Web.BlazorUI/Components/DropDown/BitDropDown.Fluent.scss
+++ b/src/Client/Web/Bit.Client.Web.BlazorUI/Components/DropDown/BitDropDown.Fluent.scss
@@ -177,7 +177,7 @@
     box-sizing: border-box;
     border-radius: 0 0 rem(2px) rem(2px);
     box-shadow: rgba(0,0,0,0.13) 0px 3.2px 7.2px 0px, rgba(0,0,0,0.11) 0px 0.6px 1.8px 0px;
-    animation-name: opacity, trans;
+    animation-name: opacity, trans-y;
     animation-duration: 0.367s;
     animation-timing-function: cubic-bezier(0.1, 0.9, 0.2, 1);
     animation-fill-mode: both;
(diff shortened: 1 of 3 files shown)


src/Client/Web/Bit.Client.Web.BlazorUI/Components/DropDown/BitDropDown.Fluent.scss
```diff
@@ -289,6 +289,41 @@
     }
 }
 
+.overflow-hidden {
+    overflow: hidden !important;
```
maryamhdr, line 293: We must never use `!important` in our codes.
plz remove this keyword and if there is some issue with removing it, you should find another way to solve it. using `!important` is not the solution :)

src/Client/Web/Bit.Client.Web.BlazorUI/Components/DropDown/BitDropDown.Fluent.scss
```diff
@@ -289,6 +289,41 @@
...
+}
+
+@media only screen and (min-width: 0) and (max-width: rem(639px)) {
```
maryamhdr, line 296: plz, use the `em` unit while defining the media query breakpoint.

src/Client/Web/Bit.Client.Web.BlazorUI/Components/DropDown/BitDropDown.Fluent.scss
```diff
@@ -289,6 +289,41 @@
...
+    .bit-drp-items-wrapper {
+        box-shadow: rgb(0 0 0 / 22%) 0 25.6px 57.6px 0, rgb(0 0 0 / 18%) 0 4.8px 14.4px 0;
+        position: fixed !important;
+        top: 0 !important;
+        left: unset !important;
+        right: 0 !important;
+        bottom: unset !important;
+        height: 100% !important;
+        width: rem(272px) !important;
```
maryamhdr, line 305: plz remove these `!important`s :)))

maryamhdr (changes requested):

src/Client/Web/Bit.Client.Web.BlazorUI/Components/DropDown/BitDropDown.Fluent.scss
RezaKargar: Here if I want to remove the `!important` from the styles I should handle it with TS/JS but its not as efficient as css media queries,
I give you an example, consider a mobile user opens a dropdown and then rotate the phone to be in the landscape mode in this case the media query will perform correctly but in TS/JS approach the user must close and re-open the dropdown to see it in proper way

msynk: in these situations, we can simply close the callout. we just need to capture the appropriate events.

msynk: we did the same in the window resize event.

msynk:
any updates? @RezaKargar

src/Client/Web/Bit.Client.Web.BlazorUI/Components/DropDown/BitDropDown.Fluent.scss
RezaKargar: But in Fluent's Demos the dropdown does not close on rotate

msynk:
I think @maryamhdr can help you here.

maryamhdr:
> There is still a problem with my solution and that is when the user rotates the phone the dropdown closes but the `overflow: hidden` style added to **body** by `overflow-hidden` class still remains, I have handled it inside the dropdown's TS/JS script not inside the window resize handler
> 
> Any ideas? @msynk @maryamhdr

Hey Reza. I need to talk to you to clarify what is the issue you are mentioning to.
How can we talk? :)

RezaKargar:
@Md23Mh has my contact,
I'll ask her to link us
@maryamhdr

msynk:
any updates?
@RezaKargar @maryamhdr

src/Client/Web/Bit.Client.Web.BlazorUI/Scripts/bit.blazorui.ts
```diff
@@ -71,7 +71,10 @@ class Bit {
 }
 
 window.addEventListener('scroll', (e: any) => {
-    if (e.target.id && Bit.currentDropDownCalloutId === e.target.id) return;
+    const minimumWidthForDropDownNormalOpen = 640;
+    if (window.innerWidth < minimumWidthForDropDownNormalOpen ||
```
maryamhdr, line 75: We must checkout 2 conditions:
In the first condition, we must check if the screen size is lower than 640px and the callout which is open is the drop-down callout, then return.
In the second one, we must check if the scroll event target is the drop-down callout, then return.

maryamhdr (changes requested):

maryamhdr (approved):

msynk (approved):

PR #1761, opened 2022-03-25
smtp username password (#1760)

### LINKED ISSUE #1760: Add feature to provide smpt server's host address, port, username and password
The todo template is using fake SMTP server during non production environments, but after that, we need to supply production ready smtp server's host address, port, username and password

mohabbati:
This closes #1760

### DIFF
--- a/src/Tooling/Bit.Tooling.Templates/TodoTemplate/Api/AppSettings.cs
+++ b/src/Tooling/Bit.Tooling.Templates/TodoTemplate/Api/AppSettings.cs
@@ -40,6 +40,6 @@
         public string Password { get; set; }
         public string DefaulFromEmail { get; set; }
         public string DefaultFromName { get; set; }
-        public bool HasCredential => !string.IsNullOrEmpty(Username) && !string.IsNullOrEmpty(Password);
+        public bool HasCredential => (string.IsNullOrEmpty(Username) is false) && (string.IsNullOrEmpty(Password) is false);
     }
 }
--- a/src/Tooling/Bit.Tooling.Templates/TodoTemplate/Api/AppSettings.cs
+++ b/src/Tooling/Bit.Tooling.Templates/TodoTemplate/Api/AppSettings.cs
@@ -36,7 +36,10 @@
     {
         public string Host { get; set; }
         public int Port { get; set; }
+        public string Username { get; set; }
+        public string Password { get; set; }
         public string DefaulFromEmail { get; set; }
         public string DefaultFromName { get; set; }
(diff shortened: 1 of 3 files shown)


src/Tooling/Bit.Tooling.Templates/TodoTemplate/Api/AppSettings.cs
```diff
@@ -36,7 +36,10 @@ public class EmailSettings
     {
         public string Host { get; set; }
         public int Port { get; set; }
+        public string Username { get; set; }
+        public string Password { get; set; }
         public string DefaulFromEmail { get; set; }
         public string DefaultFromName { get; set; }
+        public bool HasCredential => !string.IsNullOrEmpty(Username) && !string.IsNullOrEmpty(Password);
```
msynk, line 43: don't use the `not` operator (`!`) at the start of a boolean expression, it's not clear. instead, use a comparison with `false`.
for example, here it can be changed to this:
```cs
public bool HasCredential => (string.IsNullOrEmpty(Username) is false) && (string.IsNullOrEmpty(Password) is false);
```

src/Tooling/Bit.Tooling.Templates/TodoTemplate/Api/Startup/Services.cs
```diff
@@ -83,9 +83,13 @@ public static void Add(IServiceCollection services, IConfiguration configuration
...
-        services.AddFluentEmail(appSettings.EmailSettings.DefaulFromEmail, appSettings.EmailSettings.DefaultFromName)
-            .AddRazorRenderer()
-            .AddSmtpSender(appSettings.EmailSettings.Host, appSettings.EmailSettings.Port);
+        var fluentEmailServiceBuilder = services.AddFluentEmail(appSettings.EmailSettings.DefaulFromEmail, appSettings.EmailSettings.DefaultFromName)
+            .AddRazorRenderer();
+
+        if (appSettings.EmailSettings.HasCredential)
+            fluentEmailServiceBuilder.AddSmtpSender(appSettings.EmailSettings.Host, appSettings.EmailSettings.Port, appSettings.EmailSettings.Username, appSettings.EmailSettings.Password);
+        else
+            fluentEmailServiceBuilder.AddSmtpSender(appSettings.EmailSettings.Host, appSettings.EmailSettings.Port);
```
msynk, line 92: please always use `{` & `}` even for one line blocks (if, for, ...)

msynk (changes requested):

msynk:
@ysmoradi @mohabbati @maryamhdr 
the issue itself needs to be triaged.

ysmoradi (approved):

msynk (approved):

PR #1904, opened 2022-04-19
Add LabelTemplate to the `BitChoiceGroup` component options (#1848)

### LINKED ISSUE #1848: Add a new functionality to the `BitChoiceGroupOption` component like the OnRenderField property of the Fluent UI components
In the current fluent structure, each option can have an HTML next to the text.
We need to add this to the BitChoiceGroupOption class as a property of type: `RenderFragment`.

ChoiceGroup Fluent: [https://developer.microsoft.com/en-us/fluentui#/controls/web/choicegroup](url)

sajadkardel:
this closes #1848

### DIFF
--- a/src/Client/Web/Playground/Bit.Client.Web.BlazorUI.Playground/Web/Pages/Components/ChoiceGroup/BitChoiceGroupDemo.razor
+++ b/src/Client/Web/Playground/Bit.Client.Web.BlazorUI.Playground/Web/Pages/Components/ChoiceGroup/BitChoiceGroupDemo.razor
@@ -1,4 +1,5 @@
 ﻿@page "/components/choice-group"
+@using Bit.Client.Web.BlazorUI.Components.ChoiceGroup
 
 <ComponentDemo ComponentName="ChoiceGroup"
                ComponentDescription="ChoiceGroup let people select a single option from two or more choices."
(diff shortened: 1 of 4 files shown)


msynk (dismissed):

src/Client/Web/Playground/Bit.Client.Web.BlazorUI.Playground/Web/Pages/Components/ChoiceGroup/BitChoiceGroupDemo.razor
```diff
@@ -85,4 +86,62 @@
         </ExamplePreview>
     </ComponentExampleBox>
 
-</ComponentDemo>
\ No newline at end of file
+
+    <ComponentExampleBox Title="ChoiceGroup with custom options" HTMLSourceCode="@example8HtmlCode" CSharpSourceCode="@example8CSharpCode" ExampleId="example7">
```
maryamhdr, line 90: Is it example 8?
If yes, why the example id is `example7`? :D

maryamhdr (changes requested):

maryamhdr (approved):

msynk:
@Cyrus-Sushiant 
plz review this PR

msynk (approved):

src/Client/Web/Playground/Bit.Client.Web.BlazorUI.Playground/Web/Pages/Components/ChoiceGroup/BitChoiceGroupDemo.razor
```diff
@@ -85,4 +86,62 @@
...
+            <div class="example-desc">ChoiceGroup options are customized by LabelTemplate property in the ChoiceGroupOption class</div>
+            <BitChoiceGroup Label="Pick one" Options="Example8Options">
+            </BitChoiceGroup>
+        </ExamplePreview>
+    </ComponentExampleBox>
+
+</ComponentDemo>
+
+
+@code {
```
Cyrus-Sushiant, line 101: Plz, Move the code block to the `BitChoiceGroupDemo.razor.cs` file

src/Client/Web/Playground/Bit.Client.Web.BlazorUI.Playground/Web/Pages/Components/ChoiceGroup/BitChoiceGroupDemo.razor
```diff
@@ -85,4 +86,62 @@
...
+
+    // example 8
+
+    public static RenderFragment<BitChoiceGroupOption> LabelTemplateSample { get; set; }
```
Cyrus-Sushiant, line 105: Why did you use the `static` property?

Cyrus-Sushiant (changes requested):

src/Client/Web/Playground/Bit.Client.Web.BlazorUI.Playground/Web/Pages/Components/ChoiceGroup/BitChoiceGroupDemo.razor
sajadkardel: We need to write it to the code block in order to be able to set the HTML value of the RenderFragment property.
like this : [https://github.com/bitfoundation/bitframework/blob/develop/src/Client/Web/Bit.Client.Web.BlazorUI/Components/Nav/BitNav.razor](url)

src/Client/Web/Playground/Bit.Client.Web.BlazorUI.Playground/Web/Pages/Components/ChoiceGroup/BitChoiceGroupDemo.razor
sajadkardel: Untitled

This is the reason.
However, we have another solution. To change this property to a field and define it in the constructor.
what do you think? @Cyrus-Sushiant

msynk: it not importatnt, it is ok this way.

src/Client/Web/Playground/Bit.Client.Web.BlazorUI.Playground/Web/Pages/Components/ChoiceGroup/BitChoiceGroupDemo.razor
msynk: since assigning RenderFragment is very hard in a .cs file, we are forced to use the @code block in a .razor file to assign a proper value.

Cyrus-Sushiant (approved):

src/Client/Web/Playground/Bit.Client.Web.BlazorUI.Playground/Web/Pages/Components/ChoiceGroup/BitChoiceGroupDemo.razor
msynk: there is no right or wrong here. it is not an actual real situation we have here. it's a special case that we have in the demo pages that will not (and should not) occur in a real project (using such a long property initialization in a class).

PR #1945, opened 2022-04-30
Add MessageBox component (#1901)

### LINKED ISSUE #1901: Implement a simple `MessageBox` in the TodoTemplate project
In a real-world project, it's a requirement to show messages to the user. these messages could be exceptions or some other general messages that a user needs to receive.
In order to show these types of messages, we need a custom MessageBox service that can handle showing all sorts of messages (info, warning, error, ...).

msynk:
this closes #1901


msynk:
@maryamhdr & @Cyrus-Sushiant 
plz review

src/Tooling/Bit.Tooling.Templates/TodoTemplate/Web/Shared/MessageBox.razor.cs
```diff
@@ -0,0 +1,52 @@
...
+            Title = title;
+            Body = message;
+
+            StateHasChanged();
+        });
+    }
+
+    // ========================================================================
+
+    private bool IsOpen { get; set; }
```
Cyrus-Sushiant, line 34: Is the property definition of private correct? It would not be better to define a field instead of a property.

Cyrus-Sushiant (changes requested):

src/Tooling/Bit.Tooling.Templates/TodoTemplate/Web/Shared/MessageBox.razor.cs
msynk: it's a convention of the Bit Open Source projects 😁

maryamhdr (approved):

Cyrus-Sushiant (approved):

PR #1972, opened 2022-05-05
Move bgcolor & fontfamily from the reset CSS class to the MainLayout style in the `TodoTemplate` project (#1954)

### LINKED ISSUE #1954: Incorrect `background-color` and `font-family` assignment in the reset scss file of the TodoTemplate project
Currently, the `background-color` and the `font family` of the body are set in the reset file of the TodoTemplate project as shown in the following image:

image

this approach has 2 important issues:
1- the reset CSS class should not be used to set any of these visual attributes
2- setting these visual attributes for the `body` to the value of the main theme of the app is not the best practice (the body is not only for the rendering of the app)

it's better to introduce a main div container (for example in the MainLayout.razor) and set these attributes for it instead of the body.

saeedhalab:
This closes #1954


msynk (approved):

msynk:
@maryamhdr & @Cyrus-Sushiant 
plz review this PR

Cyrus-Sushiant (approved):

PR #1988, opened 2022-05-10
Resolve the `BitFileUpload` component sample issue in the `Playground` home page (#1981)

### LINKED ISSUE #1981: `BitFileUpload` component sample issue in the `playground` project home page
The `BitFileUpload` component used in the most popular section of the home page has not the proper style. It is better for the sample to fill all the available space horizontally.

It is just like this rn:
freesnippingtool com_capture_20220509172900

but it should be like this:
freesnippingtool com_capture_20220509173304

sajadkardel:
this closes #1981

### DIFF
--- a/src/Client/Web/Playground/Bit.Client.Web.BlazorUI.Playground/Web/Components/PopularComponents.razor
+++ b/src/Client/Web/Playground/Bit.Client.Web.BlazorUI.Playground/Web/Components/PopularComponents.razor
@@ -47,7 +47,8 @@
                     <BitFileUpload IsMultiSelect="true"
                            Label="Select files"
                            UploadUrl="@UploadUrl"
-                           RemoveUrl="@RemoveUrl">
+                           RemoveUrl="@RemoveUrl"
+                           Style="width:100%; max-width:none;">
                     </BitFileUpload>
                 }
                 else if (SelectedComponent.Name == "DropDown")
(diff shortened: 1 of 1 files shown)


src/Client/Web/Playground/Bit.Client.Web.BlazorUI.Playground/Web/Components/PopularComponents.razor
```diff
@@ -47,7 +47,8 @@
                     <BitFileUpload IsMultiSelect="true"
                            Label="Select files"
                            UploadUrl="@UploadUrl"
-                           RemoveUrl="@RemoveUrl">
+                           RemoveUrl="@RemoveUrl"
+                           Style="width:100%; max-width:none;">
```
msynk, line 51: since the parent element is using the flexbox, it's better to use `flex-grow:1` instead of `width:100%`
and also, can't we use a CSS class instead of an inline style?
@maryamhdr what do you think?

msynk (changes requested):

src/Client/Web/Playground/Bit.Client.Web.BlazorUI.Playground/Web/Components/PopularComponents.razor
msynk: yeah, you are right about the `max-width`, but my concern is about the `width` property.

msynk: and about the inline style, I'm ok with using it if there are some other cases on the same page.
@maryamhdr can enlighten us :)

msynk (approved):

msynk:
@maryamhdr plz review

maryamhdr (approved):

PR #1997, opened 2022-05-11
Add login form in swagger UI

### LINKED ISSUE #1798: Enhance Swagger to prompt a login form for secured APIs of the TodoTemplate project
Currently, it is required to provide an access token in order to call the secured APIs of the template project which is hard to use. it is better to config the swagger to show a login form with a username & password field in order to do a login process before calling these APIs.

milad-karimifard:
This closes #1798


ysmoradi (changes requested):
I think we've lost the chance to provide a sample to our audience to show how they can call the api provided by todo template using javascript.
I'd prefer using async/await + fetch. I also recommend using const/let instead of var.
This code need heavy refactor to make sure it uses modern javascript
I also prefer another name other than customer-swagger-login.js, something more general is better in my opinion. For example swagger-utils.js What do you think @msynk @milad-karimifard

msynk:
@milad-karimifard 
first of all, thanks a lot for your contribution. I appreciate it a lot. 👍
the code you've provided seems to be borrowed from somewhere like this article:
https://www.codedesigntips.com/2021/06/28/swagger-ui-with-login-form-and-role-based-api-visibility/
so if it's borrowed plz provide the source in the code.
and also the javascript code needs to be converted to typescript first before getting reviewed.

msynk:
IMO, it seems fine now.
@ysmoradi could you plz review this PR and also test the script on your machine?

msynk (approved):

ysmoradi (changes requested):

src/Tooling/Bit.Tooling.Templates/TodoTemplate/Api/wwwroot/swagger/swagger-utils.js
```diff
@@ -0,0 +1,181 @@
...
+    const loginButton = document.createElement("button")
+    loginButton.type = "submit";
+    loginButton.type = "button";
+    loginButton.classList.add("btn");
+    loginButton.classList.add("auth");
+    loginButton.classList.add("authorize");
+    loginButton.classList.add("button");
+    loginButton.innerText = "Login";
+    loginButton.onclick = function () {
+        let userName = userNameInput.value;
```
ysmoradi, line 103: Use const instead of let when possible

src/Tooling/Bit.Tooling.Templates/TodoTemplate/Api/wwwroot/swagger/swagger-utils.js
```diff
@@ -0,0 +1,181 @@
...
+            },
+        };
+        swagger.authActions.authorize(obj);
+    } else{
+        alert(await response.text())
+    }
+}
+
+const attachReloadPageToEvent = (swagger, caller, args) => {
+    const result = caller(args);
```
ysmoradi, line 146: use `async await`
```js
const result = await caller(args);
reloadPage(swagger);
```

ysmoradi (approved):

ysmoradi:
I'm gonna merge this PR right after testing that on my local machine @milad-karimifard @msynk 
Thanks!

PR #2027, opened 2022-05-15
Fix error in `BitRadioButtonList` when selecting an item #1998

### LINKED ISSUE #1998: Get error in `BitRadioButtonList` when selecting an item
In `BitRadioButtonList`, assigning value to a Tuple object throws an error when selecting an item, because it can not convert string to Tuple.

image

Cyrus-Sushiant:
This closes #1998

### DIFF
--- a/src/Client/Web/Bit.Client.Web.BlazorUI/Extensions/ObjectExtensions.cs
+++ b/src/Client/Web/Bit.Client.Web.BlazorUI/Extensions/ObjectExtensions.cs
@@ -5,11 +5,6 @@ namespace Bit.Client.Web.BlazorUI
 {
     public static class ObjectExtensions
     {
-        public static string? GetValueAsStringFromProperty(this object? obj, string propertyName)
-        {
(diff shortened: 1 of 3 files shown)


src/Client/Web/Bit.Client.Web.BlazorUI/Extensions/ObjectExtensions.cs
```diff
@@ -5,9 +5,14 @@ namespace Bit.Client.Web.BlazorUI
 {
     public static class ObjectExtensions
     {
-        public static string? GetValueFromProperty(this object? obj, string propertyName)
+        public static string? GetValueAsStringFromProperty(this object? obj, string propertyName)
         {
-            return obj?.GetType().GetProperty(propertyName)?.GetValue(obj)?.ToString();
+            return obj?.GetValueAsObjectFromProperty(propertyName)?.ToString();
+        }
```
msynk, line 11: remove this redundant method and instead use the ToString method at the caller site directly.

msynk (approved):

PR #2030, opened 2022-05-16
Remove old format fonts (#2029)

### LINKED ISSUE #2029: Bit components are using old format fonts
Based on https://caniuse.com/woff2
We can use only the woff2 format of the fonts that we need

ysmoradi:
This closes #2029


msynk (approved):

msynk:
@maryamhdr plz review this PR

PR #2066, opened 2022-05-22
Fixed health check issues (#2065)

### LINKED ISSUE #2065: todo template aspnetcore healthchecks issues
1- `Health checks` service registration is not useable in `health checks ui`. It also has a name that broke our naming convention.
So, Instead of
```cs
setup.AddHealthCheckEndpoint("BitHealthCheck", "/healthz");
```
We've to use something like
```cs
setup.AddHealthCheckEndpoint("TodoHealthChecks", webHostEnvironment.IsDevelopment() ? "https://localhost:5001/healthz" : "/healthz");
```
I'd recommend using `TodoHealthChecks` instead of `BitHealthCheck`
Note that an absolute URL is needed to make health checks UI work
(https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/issues/410)

2- Health checks `UI` was not even registered! (This is not working for sure and I've no idea how it was tested before)

image

3- `Health checks` gets added to the pipeline, no matter if it's disabled or not in-app configuration, so I'd recommend checking if health checks is enabled or not using configuration, then we can add it to the aspnet core pipeline

image

4- `health checks ui` is unable to call `/healthz`, so let's add `/healthz` to PWA `serverHandledUrls`

ysmoradi:
This closes #2065

### DIFF
--- a/src/Tooling/Bit.Tooling.Templates/TodoTemplate/Api/Startup/Middlewares.cs
+++ b/src/Tooling/Bit.Tooling.Templates/TodoTemplate/Api/Startup/Middlewares.cs
@@ -7,7 +7,7 @@ namespace TodoTemplate.Api.Startup
 {
     public class Middlewares
     {
-        public static void Use(IApplicationBuilder app, IHostEnvironment env)
+        public static void Use(IApplicationBuilder app, IHostEnvironment env, IConfiguration configuration)
         {
             if (env.IsDevelopment())
             {
@@ -31,7 +31,7 @@ namespace TodoTemplate.Api.Startup
             {
                 options.InjectJavascript("/swagger/swagger-utils.js");
             });
-            
+
             app.UseResponseCompression();
 
             app.UseStaticFiles(new StaticFileOptions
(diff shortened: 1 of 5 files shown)


src/Tooling/Bit.Tooling.Templates/TodoTemplate/Api/Startup/Middlewares.cs
```diff
@@ -61,10 +61,20 @@ public static void Use(IApplicationBuilder app, IHostEnvironment env)
             app.UseEndpoints(endpoints =>
             {
                 endpoints.MapControllers().RequireAuthorization();
-                endpoints.MapHealthChecks("/healthz", new HealthCheckOptions
+
+                var appsettings = configuration.GetSection(nameof(AppSettings)).Get<AppSettings>();
+
+                var healthCheckSettings = appsettings.HealCheckSettings;
+
+                if (healthCheckSettings.EnableHealthChecks is true)
```
msynk, line 69: please remove the redundant `is true` part in this line

msynk (approved):

msynk (approved):

PR #2170, opened 2022-06-08
Resolve the issue of allowing form submission by pressing Enter without checking the privacy policy on the Sign-Up page (#2158)

### LINKED ISSUE #2158: Sigup without check the privacy police
On the  SignUp page without checking "I agree to the Bit police", I could post the form to the server by pressing the Enter key.

Cyrus-Sushiant:
This closes #2158

### DIFF
--- a/src/Tooling/Bit.Tooling.Templates/TodoTemplate/Web/Pages/SignUp.razor.cs
+++ b/src/Tooling/Bit.Tooling.Templates/TodoTemplate/Web/Pages/SignUp.razor.cs
@@ -26,7 +26,7 @@ public partial class SignUp
 
     private async Task DoSignUp()
     {
-        if (IsLoading || SignUpModel.IsAcceptPrivacy is false)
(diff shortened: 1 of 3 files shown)


maryamhdr (dismissed):

src/Tooling/Bit.Tooling.Templates/TodoTemplate/Web/Pages/SignUp.razor.cs
```diff
@@ -26,7 +26,7 @@ public partial class SignUp
 
     private async Task DoSignUp()
     {
-        if (IsLoading)
+        if (IsLoading || SignUpModel.IsAcceptPrivacy is false)
```
msynk, line 29: this is like a workaround. I don't like it.
shouldn't we try to validate the form instead?

msynk (changes requested):

msynk (approved):

PR #2202, opened 2022-06-15
Resolve issue two-way binding parameters in BitSlider ranged (#2181)

### LINKED ISSUE #2181: BitSlider with Ranged true doesn't set values
**Describe the bug**
When using BitSldier binding, the values do not set properly.
`<BitSlider
             @bind-LowerValue=lower
             @bind-UpperValue=upper
             Min="1"
             Max="80"
             Step="1"
             IsRanged="true"></BitSlider>`

inside code:
`public double? lower;
public double? upper;`

When you pull on the slider, it counts up, but when releasing it resets to initial value. Have tried with no value, or a set value but anytime I move the slider, it just resets to the initial value.

I removed the IsRanged settings and the Upper value and it works as intended.

**To Reproduce**
I have tried in multiple different ways to get this to work, every time I get the same result. To reproduce, just created a blank new blazor server app and it still does the same. If I create a button that sets the values then the slider moves as it should, but sliding it doesn't update the variable that is bound.

**Expected behavior**
When moving a slider, it should update the bound value.

**Bit Blazor client (If applicable)**
Bit.Client.Web.BlazorUI version 11.0.0 and tried 12.0.0-pre-02
Blazor Server Net 6
Chrome 102.0.5005.63

Cyrus-Sushiant:
This closes #2181

### DIFF
--- a/src/Client/Web/Bit.Client.Web.BlazorUI/Components/Slider/BitSlider.razor.cs
+++ b/src/Client/Web/Bit.Client.Web.BlazorUI/Components/Slider/BitSlider.razor.cs
@@ -236,12 +236,12 @@ namespace Bit.Client.Web.BlazorUI
         {
             if (DefaultUpperValue.HasValue)
             {
-                upperValue = DefaultUpperValue.Value;
+                UpperValue = DefaultUpperValue.Value;
             }
 
             if (DefaultLowerValue.HasValue)
             {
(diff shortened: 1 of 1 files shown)


msynk:
@Cyrus-Sushiant 
plz update this PR based on our conversation

msynk (approved):

PR #2212, opened 2022-06-16
Migrate HttpResponseExceptionFilter to HttpResponseExceptionMiddleware (#2174)

### LINKED ISSUE #2174: HttpResponseExceptionFilter is not working in all scenario
The HttpResponseExceptionFilter is an action filter, so it runs if only all action filters before that run correctly and don't cut to circuits (return a response of request and finish request). 
The problem is if an exception happens in another action filter, we can't be able to catch that and return our specific payload. On the other hand, because this action filter does not apply to all requests, we can't set the `Request-Id` header for all requests.

I suggest replacing the action filter with middleware so we can mark all requests and catch-all exceptions. if we do that, we are capable of logging all unknown exceptions easily and globally in a single place

milad-karimifard:
this closes #2174


src/Tooling/Bit.Tooling.Templates/TodoTemplate/Api/Middlewares/HttpResponseExceptionHandlerMiddleware.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            if (exp is TargetInvocationException)
+                return exp.InnerException ?? exp;
+
+            return exp;
+        }
+    }
+}
+
+public static class HttpResponseExceptionHandlerMiddlewareExtensions
```
msynk, line 54: please move this class to a separate file.
and check out other Extension classes that are already implemented in the projects.

src/Tooling/Bit.Tooling.Templates/TodoTemplate/Api/Middlewares/HttpResponseExceptionHandlerMiddleware.cs
```diff
@@ -0,0 +1,60 @@
...
+            context.Response.StatusCode = statusCode;
+            await context.Response.WriteAsJsonAsync(restExceptionPayload);
+        }
+
+        Exception UnWrapException(Exception exp)
+        {
+            if (exp is TargetInvocationException)
+                return exp.InnerException ?? exp;
+
+            return exp;
```
msynk, line 49: please use a ternary operator instead of the if statement, something like this:

```cs
return (exp is TargetInvocationException && exp.InnerException is not null) 
           ? exp.InnerException
           : exp
```

src/Tooling/Bit.Tooling.Templates/TodoTemplate/Api/Middlewares/HttpResponseExceptionHandlerMiddleware.cs
```diff
@@ -0,0 +1,60 @@
...
+            if (exception is ResourceValidationException validationException)
+            {
+                restExceptionPayload.Details = validationException.Details;
+            }
+
+            context.Response.StatusCode = statusCode;
+            await context.Response.WriteAsJsonAsync(restExceptionPayload);
+        }
+
+        Exception UnWrapException(Exception exp)
```
msynk, line 44: please always use the access modifer. (`private` here)

src/Tooling/Bit.Tooling.Templates/TodoTemplate/Api/Middlewares/HttpResponseExceptionHandlerMiddleware.cs
```diff
@@ -0,0 +1,60 @@
...
+            RestExceptionPayload restExceptionPayload = new RestExceptionPayload
+            {
+                Message = isKnownException || webHostEnvironment.IsDevelopment() ? exception.Message : nameof(UnknownException),
+                ExceptionType = isKnownException ? exception.GetType().FullName : typeof(UnknownException).FullName
+            };
+
+            if (exception is ResourceValidationException validationException)
+            {
+                restExceptionPayload.Details = validationException.Details;
+            }
```
msynk, line 38: please extract a method to create the `restExceptionPayload` in order to make the `Invoke` method simpler and easier to read.

msynk (changes requested):

src/Tooling/Bit.Tooling.Templates/TodoTemplate/Api/Middlewares/HttpResponseExceptionHandlerMiddleware.cs
```diff
@@ -0,0 +1,58 @@
...
+            var statusCode = (int)(exception is RestException restExp ? restExp.StatusCode : HttpStatusCode.InternalServerError);
+            
+            var restExceptionPayload = RestExceptionPayloadBuilder(webHostEnvironment, isKnownException, exception);
+
+            context.Response.StatusCode = statusCode;
+            await context.Response.WriteAsJsonAsync(restExceptionPayload);
+        }
+    }
+
+    private RestExceptionPayload RestExceptionPayloadBuilder(IHostEnvironment webHostEnvironment, bool isKnownException, Exception exception)
```
msynk, line 36: please rename this method to `CreateRestExceptionPayload`.
this name is not a suitable name for a method. it's rather a class name.

msynk (changes requested):

ysmoradi (approved):

msynk (approved):

PR #2238, opened 2022-06-22
Fix the incorrect behavior of the parameters in the BitNumricTextField (#2236)

### LINKED ISSUE #2236: BitTextNumricFiled doesn't act as expected because the default values are different from what was set

### Describe the bug

When I use `BitTextNumricFiled` component, the behavior produced to me is not what I expected. What happened is the parameters I don't set get their own default values, but the problem is the default values are incorrect. For example, If I don't set `min` or `max` value, their set to `0`. also, because both parameters are set to zero, any value I try to set becomes zero because it is not in the supported range

### Expected Behavior

I expect if I don't set values for parameters, everything works well.

### Steps To Reproduce

_No response_

### Exceptions (if any)

_No response_

### .NET Version

6.0.300

### Anything else?

_No response_

Cyrus-Sushiant:
This closes #2236


src/Client/Web/Bit.Client.Web.BlazorUI/Components/NumericTextField/BitNumericTextField.razor.cs
```diff
@@ -113,16 +126,20 @@ public BitNumericTextField()
...
         /// </summary>
         [Parameter]
         public TValue? Step
         {
-            get => GetGenericValue(step);
-            set => step = GetDoubleValueOrDefault(value, 1);
+            get => step;
+            set
+            {
+                internalStep = GetDoubleValueOrDefault(value, 1)!.Value;
```
msynk, line 140: can't you change this line to this?
```cs
internalStep = GetDoubleValueOrDefault(value) ?? 1
```

src/Client/Web/Bit.Client.Web.BlazorUI/Components/NumericTextField/BitNumericTextField.razor.cs
```diff
@@ -623,7 +639,7 @@ protected override void Dispose(bool disposing)
         }
 
         /// <inheritdoc />
-        protected override bool TryParseValueFromString(string? value, [MaybeNullWhen(false)] out TValue? result, [NotNullWhen(false)] out string? validationErrorMessage)
+        protected override bool TryParseValueFromString(string? value, [MaybeNullWhen(false)] out TValue result, [NotNullWhen(false)] out string? validationErrorMessage)
```
msynk, line 642: this change should be revoked, the result needs to be nullable or there will be a permanent warning:

`Parameter 'result' must have a non-null value when exiting with 'true'.`

msynk (approved):

PR #2276, opened 2022-06-26
Fix the CalculatePrecision method in BitNumericTextField #2274

### LINKED ISSUE #2274: BitNumericTextField throw exception on some scenarios when value type is float
When I set `BitNumericTextField` parameters with the below values, an exception is thrown, and my page doesn't render correctly.

Min: 0f,
Max: 100f,
Step: 0.1f,
DefaultValue: 25.68f

Exception:
System.ArgumentOutOfRangeException: Rounding digits must be between 0 and 15, inclusive. (Parameter 'digits')

Cyrus-Sushiant:
This closes #2274

### DIFF
--- a/src/Client/Web/Bit.Client.Web.BlazorUI/Components/NumericTextField/BitNumericTextField.razor.cs
+++ b/src/Client/Web/Bit.Client.Web.BlazorUI/Components/NumericTextField/BitNumericTextField.razor.cs
@@ -434,6 +434,8 @@ namespace Bit.Client.Web.BlazorUI
 
         private int CalculatePrecision(TValue? value)
         {
+            if (value is null) return 0;
+
             var pattern = isDecimals ? @"[1-9]([0]+$)|\.([0-9]*)" : @"(^-\d+$)|\d+";
             var regex = new Regex(pattern);
             if (regex.IsMatch($"{value}") is false) return 0;
--- a/src/Client/Web/Bit.Client.Web.BlazorUI/Components/NumericTextField/BitNumericTextField.razor.cs
+++ b/src/Client/Web/Bit.Client.Web.BlazorUI/Components/NumericTextField/BitNumericTextField.razor.cs
@@ -250,7 +250,7 @@ namespace Bit.Client.Web.BlazorUI
                 internalMin -= internalMax;
             }
 
-            precision = Precision is not null ? Precision.Value : CalculatePrecision(internalStep);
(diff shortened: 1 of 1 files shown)


src/Client/Web/Bit.Client.Web.BlazorUI/Components/NumericTextField/BitNumericTextField.razor.cs
```diff
@@ -432,13 +432,13 @@ private async Task HandleFocus(FocusEventArgs e)
             }
         }
 
-        private int CalculatePrecision(double value)
+        private int CalculatePrecision(TValue? value)
         {
             var pattern = isDecimals ? @"[1-9]([0]+$)|\.([0-9]*)" : @"(^-\d+$)|\d+";
             var regex = new Regex(pattern);
-            if (regex.IsMatch(value.ToString(CultureInfo.InvariantCulture)) is false) return 0;
+            if (regex.IsMatch($"{value}") is false) return 0;
```
msynk, line 439: what is this change for?
I think the previous version was much better.

src/Client/Web/Bit.Client.Web.BlazorUI/Components/NumericTextField/BitNumericTextField.razor.cs
```diff
@@ -432,13 +432,13 @@ private async Task HandleFocus(FocusEventArgs e)
             }
         }
 
-        private int CalculatePrecision(double value)
+        private int CalculatePrecision(TValue? value)
```
msynk, line 435: shouldn't the nullability of the value be checked first?

msynk (changes requested):

src/Client/Web/Bit.Client.Web.BlazorUI/Components/NumericTextField/BitNumericTextField.razor.cs
Cyrus-Sushiant: because the value is a generic type and is not support ToString wite `CultureInfo.InvariantCulture` argument.

msynk: oh right,
the next version of C# will resolve these types of issues with static abstract members for interfaces.

msynk (approved):

PR #2287, opened 2022-06-27
Change the namespace of the `AutoInject` attribute (#2286)

### LINKED ISSUE #2286: Move the `AutoInject` attribute class to the `Microsoft.Extensions.DependencyInjection` namespace
Currently, the `AutoInject` attribute is defined in a class in the `Bit.Tooling.SourceGenerators` namespace which is incorrect. this class should be moved to the `Microsoft.Extensions.DependencyInjection` namespace.

msynk:
this closes #2286

### DIFF
--- a/src/Tooling/Bit.Tooling.SourceGenerators/AutoInject/AutoInjectAttribute.cs
+++ b/src/Tooling/Bit.Tooling.SourceGenerators/AutoInject/AutoInjectAttribute.cs
@@ -1,6 +1,6 @@
 ﻿using System;
 
-namespace Bit.Tooling.SourceGenerators;
+namespace Microsoft.Extensions.DependencyInjection;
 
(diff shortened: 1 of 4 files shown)


ysmoradi (approved):

src/Tooling/Bit.Tooling.SourceGenerators/AutoInject/AutoInjectAttribute.cs
```diff
@@ -1,6 +1,6 @@
 ﻿using System;
 
-namespace Bit.Tooling.SourceGenerators;
+namespace Microsoft.Extensions.DependencyInjection;
```
ysmoradi, line 3: I'd prefer ms.ext.di.abstractions, because our approach works well with all implementations of ms.ext.di.abstractions such as autofac.ext.di

ysmoradi (approved):

PR #2303, opened 2022-06-28
Fix the not disabled dates issue on the BitDatePicker based on min and max parameters (#2246)

### LINKED ISSUE #2246: Not disabled dates on the `BitDatePicker` based on `Min` and `Max` parameters
Untitled

In the image above, the amount of Max date parameter is equal to July 23
As you can see, it's disabled from July 23 to July 31 and at the start of the new month is enabled again...

It is expected to be disabled from July 23 at any time in the future...
(Note that the issue also exists for the Min date)

sajadkardel:
this closes #2246


sajadkardel:
I forked the project again due to structural changes in the project. For this reason, this PR, which was sent from the previous fork, was closed automatically.

PR #2312, opened 2022-06-29
Fix disabled dates issues of Min & Max parameters in the `BitDatePicker` component (#2246)

### LINKED ISSUE #2246: Not disabled dates on the `BitDatePicker` based on `Min` and `Max` parameters

sajadkardel:
this close #2246


msynk:
@sajadkardel 
plz resolve the conflict

msynk (dismissed):

sajadkardel:
> @sajadkardel plz resolve the conflict

Resolved 💯

msynk (approved):

PR #2356, opened 2022-07-02
Accept & process non-partial classes in AutoInject (#2347)

### LINKED ISSUE #2347: The `AutoInject` source generator ignores non-partial classes
The `AutoInject` source generator should not ignore non-partial classes. 
It should always generate the source code for all occurrences of the attribute.

milad-karimifard:
this closes #2347

### DIFF
--- a/src/SourceGenerators/Bit.SourceGenerators/AutoInject/AutoInjectSourceGenerator.cs
+++ b/src/SourceGenerators/Bit.SourceGenerators/AutoInject/AutoInjectSourceGenerator.cs
@@ -14,7 +14,7 @@ namespace Bit.SourceGenerators;
 public class AutoInjectSourceGenerator : ISourceGenerator
 {
     private static readonly DiagnosticDescriptor NonPartialClassError = new DiagnosticDescriptor(id: "BITGEN001",
-                                                                                              title: "The class use AutoInject must be partial",
+                                                                                              title: "The class needs to be partial",
                                                                                               messageFormat: "{0} is not partial. The AutoInject attribute needs to be used only in partial classes.",
                                                                                               category: "Bit.SourceGenerators",
                                                                                               DiagnosticSeverity.Error,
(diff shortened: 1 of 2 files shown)


src/SourceGenerators/Bit.SourceGenerators/AutoInject/AutoInjectSourceGenerator.cs
```diff
@@ -4,13 +4,22 @@
...
+using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Text;
 
 namespace Bit.SourceGenerators;
 
 [Generator]
 public class AutoInjectSourceGenerator : ISourceGenerator
 {
+    private static readonly DiagnosticDescriptor NonPartialClassError = new DiagnosticDescriptor(id: "BITGEN001",
+                                                                                              title: "The class use AutoInject must be partial",
```
ysmoradi, line 17: This message has nothing to say about classes which their base class has auto injected member
@msynk Could you please provide better error message?

ysmoradi (changes requested):

src/SourceGenerators/Bit.SourceGenerators/AutoInject/AutoInjectSourceGenerator.cs
msynk: sure, here:

`{CLASS_NAME} is not partial. The AutoInject attribute needs to be used only in partial classes.`

remember to extract the correct class name (for example if needed the base class name should be used).

ysmoradi (approved):

src/SourceGenerators/Bit.SourceGenerators/AutoInject/AutoInjectSourceGenerator.cs
msynk, line 17: plz change this title to `The class needs to be partial`

msynk (changes requested):

msynk (approved):

PR #2369, opened 2022-07-03
The `TodoTemplate` cleanup (#2343)

### LINKED ISSUE #2343: TodoTemplate project Cleanup
Currently, there are some misuses of the name `TodoTemplate` or `todoTemplate` in the projects of the TodoTemplate which creates some issues in the generated projects using the Project Template. it is required to change these names to other appropriate replacements.

msynk:
this closes #2343


src/Templates/TodoTemplate/Bit.TodoTemplate/App/TodoTemplate.App.csproj
```diff
@@ -12,7 +12,7 @@
         <ApplicationTitle>TodoTemplate</ApplicationTitle>
 
         <!-- App Identifier -->
-        <ApplicationId>com.bitplatform.TodoTemplate</ApplicationId>
+        <ApplicationId>com.bitplatform.Todo.Template</ApplicationId>
```
ysmoradi, line 15: Why? @msynk

msynk: if we don't change it, the `TodoTemplate` is going to be replaced with the target project name which is meaningless:
`com.bitplatform.testApp`

src/Templates/TodoTemplate/Bit.TodoTemplate/Web/Services/Implementations/AuthenticationService.cs
```diff
@@ -0,0 +1,38 @@
...
+
+public class AuthenticationService : IAuthenticationService
+{
+    private readonly HttpClient _httpClient;
+
+    private readonly IJSRuntime _jsRuntime;
+
+    private readonly AppAuthenticationStateProvider _authenticationStateProvider;
+
+    public AuthenticationService(HttpClient httpClient, IJSRuntime jsRuntime,
```
ysmoradi, line 13: Shall we use AutoInject here? @msynk

ysmoradi (dismissed):

src/Templates/TodoTemplate/Bit.TodoTemplate/Web/Services/Implementations/AuthenticationService.cs
msynk: sure

ysmoradi (approved):

PR #2371, opened 2022-07-03
Add Generic Class Support in AutoInject (#2300)

### LINKED ISSUE #2300: `AutoInject` does not support `Generic` classes
When you try to add `AutoInject` in a `Generic` class then the Source Generator generates a `Non-Generic` class.

For instance, `BitNumericTextField` is a `Generic` class:
```C#
public partial class BitNumericTextField<TValue>
{
    //
}
````

The result of the Source Generator is:

```C#
public partial class BitNumericTextField
{        
    [Inject]
    [EditorBrowsable(EditorBrowsableState.Never)]
    private Microsoft.JSInterop.IJSRuntime ____JsRuntime { get => _jsRuntime; set => _jsRuntime = value; }
}
```

milad-karimifard:
this closes #2300

### DIFF
--- a/src/SourceGenerators/Bit.SourceGenerators/AutoInject/AutoInjectNormalClassHandler.cs
+++ b/src/SourceGenerators/Bit.SourceGenerators/AutoInject/AutoInjectNormalClassHandler.cs
@@ -33,35 +33,12 @@ namespace {classNamespace}
     [global::System.CodeDom.Compiler.GeneratedCode(""Bit.SourceGenerators"",""{AutoInjectHelper.GetPackageVersion()}"")]
     [global::System.Diagnostics.DebuggerNonUserCode]
     [global::System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage]
-    public partial class {GenerateClassName(classSymbol)}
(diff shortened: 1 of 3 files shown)


ysmoradi (dismissed):

src/SourceGenerators/Bit.SourceGenerators/AutoInject/AutoInjectNormalClassHandler.cs
```diff
@@ -33,14 +33,37 @@ namespace {classNamespace}
...
+                sb.Append($"{item.Name},");
+            }
+
+            sb.Length--;
+            sb.Append(">");
+
+            genericTypesName = sb.ToString();
+        }
+
+        return $"{classSymbol.Name}{genericTypesName}";
```
msynk, line 64: please change this method's implementation based on a similar method in the `ComponentDetailsController` of the Bit.BlazorUI Playground (the GetTypeName method in the controller).

msynk (changes requested):

ysmoradi (approved):

msynk (approved):

PR #2513, opened 2022-07-13
Highlight featured components of the `Bit.BlazorUI` in the Bit BlazorUI demo project (#2458)

### LINKED ISSUE #2458: Highlight featured components of the `Bit.BlazorUI` in the Bit `BlazorUI` demo project
It is better to highlight the most featured components of the `Bit.BlazorUI` in the component list presented on the home page of the  Bit `BlazorUI` demo project. For example, use a label named `Featured` in front of each target component.
Here is the list of featured components:
- BitDataGrid
- BitFileUpload
- BitColorPicker
- BitDatePicker
- BitDropDown
- BitNav

sajadkardel:
this closes #2458

### DIFF
--- a/src/BlazorUI/Playground/Bit.BlazorUI.Playground/Web/Styles/styles.scss
+++ b/src/BlazorUI/Playground/Bit.BlazorUI.Playground/Web/Styles/styles.scss
@@ -75,6 +75,8 @@ a {
     cursor: pointer;
 
     a {
+        display: flex;
+        align-items: center;
         color: inherit;
         font-family: inherit;
 
--- a/src/BlazorUI/Playground/Bit.BlazorUI.Playground/Web/Pages/HomePage.razor
+++ b/src/BlazorUI/Playground/Bit.BlazorUI.Playground/Web/Pages/HomePage.razor
@@ -97,56 +97,121 @@
             <div class="list-col">
                 <div class="list">
                     <div class="list-title">Basic inputs</div>
-                    <BitLink Class="component-list-item" Href="/components/button" title="Button">Button</BitLink>
-                    <BitLink Class="component-list-item" Href="/components/action-button" title="ActionButton">ActionButton</BitLink>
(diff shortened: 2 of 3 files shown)


sajadkardel:
output:

Untitled

src/BlazorUI/Playground/Bit.BlazorUI.Playground/Web/Styles/styles.scss
```diff
@@ -75,6 +75,8 @@ a {
     cursor: pointer;
 
     a {
+        display: flex;
+        align-items: center;
         color: inherit;
         font-family: inherit;
```
msynk, line 82: since this class (component-list-item) is only used for the HomePage components, shouldn't we move it to the `HomePage.razor.scss` file?

msynk (changes requested):

src/BlazorUI/Playground/Bit.BlazorUI.Playground/Web/Styles/styles.scss
sajadkardel: As a convention, we moved all CSS of bit component classes to the `style.scss` file, regardless of whether it is for a specific page.
I think this is the reason.
What do you think? @maryamhdr

maryamhdr: @sajadkardel 
Could you check if the `component-list-item` class be declared in the `HomePage.razor.scss` file, the items get the correct styles or not?

msynk: @sajadkardel 
no we didn't.
we moved some classes but the reason is not what you explained. plz test it and share the result with us.
but I think since the isolated CSS classes have issues with components, I think it won't work for this situation.

sajadkardel: I tested it. didn't work in the `HomePage.razor.scss` file @maryamhdr @msynk

msynk (approved):

PR #2572, opened 2022-07-16
Changed the public access modifier to internal in all extension classes (#2086)

### LINKED ISSUE #2086: Change the access modifier in the extension classes
Change the access modifier in the extension classes, because it should not be accessible to the classes and their methods outside the components.

mohabbati:
This closes #2086


src/BlazorUI/Playground/Bit.BlazorUI.Playground/Web/Extensions/EnumExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+            }
+            else if (bitIconName == default)
+            {
+                return null;
+            }
+        }
+
+        return bitIconName.GetDisplayName();
+    }
+}
```
msynk, line 79: instead of redefining this class use `[assembly: InternalsVisibleTo("Bit.BlazorUI.Tests")]`

src/BlazorUI/Playground/Bit.BlazorUI.Playground/Web/Extensions/StringExtensions.cs
```diff
@@ -0,0 +1,18 @@
...
+            : string.IsNullOrEmpty(value) is false;
+    }
+
+    public static bool HasNoValue(this string? value, bool ignoreWhiteSpace = true)
+    {
+        return ignoreWhiteSpace
+            ? string.IsNullOrWhiteSpace(value)
+            : string.IsNullOrEmpty(value);
+    }
+}
```
msynk, line 18: again, instead of redefining this class use `[assembly: InternalsVisibleTo("Bit.BlazorUI.Tests")]`

msynk (changes requested):

msynk:
@mohabbati 
did you try to build the project and run the tests?
I'm pretty sure that some of these extension methods are also used in the Tests project.

mohabbati:
> Tests project.

All projects in the `Bit.BlazorUI` solution were built successfully. And 2384 tests passed after running tests.

msynk:
you are right, there is already an `InternalsVisibleTo("Bit.BlazorUI.Tests")` set in the BitColor.cs file.

src/BlazorUI/Bit.BlazorUI/Extensions/StringExtensions.cs
```diff
@@ -1,15 +1,18 @@
-﻿namespace Bit.BlazorUI;
+﻿using System.Runtime.CompilerServices;
 
-public static class StringExtensions
+[assembly: InternalsVisibleTo("Bit.BlazorUI.Playground.Web")]
```
msynk, line 3: since this is an assembly level attribute, you need to apply it once.
please remove this duplicate line.

msynk (changes requested):

msynk (approved):

PR #2576, opened 2022-07-17
Fix enabling submit button in `SignUp` with considering correct validation for `TodoTemplate` (#2488)

### LINKED ISSUE #2488: The sign up button gets enabled style while there is some invalid field in the `TodoTemplate` project
I entered the sign-up page and filled all the fields, then checked the privacy checkbox. After these, the sign-in button became enabled. 
Then I removed some characters from the password/email...But the styles of the sign-up button still remained enabled.
Notice that even in this state, the sign-up button didn't work and the sign-up request hadn't been sent.

freesnippingtool com_capture_20220712141827
freesnippingtool com_capture_20220712142724
freesnippingtool com_capture_20220712142802

milad-karimifard:
this closes #2488


msynk:

PR #2577, opened 2022-07-17
Fix enabling submit button in `ForgetPassword` with considering correct validation for `TodoTemplate` (#2492)

### LINKED ISSUE #2492: The submit button gets enabled style while there is an invalid field in the forget password page of the `TodoTemplate` project
I entered the forget password page and filled the email field. After that, the submit button became enabled.
Then I removed some characters from the email...But the styles of the button still remained enabled.
Notice that even in this state, the submit button didn't work and the reset password request hadn't been sent.

freesnippingtool com_capture_20220712145721

milad-karimifard:
this closes #2492


msynk:

PR #2578, opened 2022-07-17
Fix enabling submit button in `ResetPassword` with considering correct validation for `TodoTemplate` (#2494)

### LINKED ISSUE #2494: The button gets enabled style while there is some invalid field in the reset password page of the `TodoTemplate` project
I entered the reset password page via the reset link and filled all the fields. After that, the button became enabled.
Then I removed some characters from the fields...But the styles of the button still remained enabled.
Notice that even in this state, the button didn't work and the reset password request hadn't been sent.

freesnippingtool com_capture_20220712150237

milad-karimifard:
this closes #2494


msynk:

PR #2653, opened 2022-07-21
Save the version of the project templates (#2408)

### LINKED ISSUE #2408: Save the version of the template installed in the project.

### Is your feature request related to a problem? Please describe the problem.

It seems that you should save the version of each template that is installed and used in the projects somewhere so that the problems of the projects that use the template can be checked.

### Describe the solution you'd like

I have no idea where to save this version, but for example, we can write it somewhere like above the program class and indicate that that line should not be deleted for future checks. Or, create a file with CodeGenerator and write it in it.

### Additional context

_No response_

MeysamBahadori:
This closes #2408


msynk:
what is this PR? 🧐

msynk (approved):

ysmoradi (approved):

PR #2668, opened 2022-07-22
Fix implement two-way binding in the BitRadioButtonGroup #2541

### LINKED ISSUE #2541: Two-way binding on the `BitRadioButtonGroup` does not work
Two-way biding on BitRadioButtonGroup does not work
#### It does not work on
- First sample of [Component site](https://components.bitplatform.dev/components/radio-button-group)
- My code
```
<div>
    <BitTextField @bind-Value="MyValue" Placeholder="Select one of A, B or C"></BitTextField>
<span>Or</span>
    <BitButton  OnClick="@(()=>MyValue="A")">Select A</BitButton>
</div>

<BitRadioButtonGroup Name="Group1" Label="Pick one" IsRequired="true" @bind-Value="MyValue">
    <BitRadioButtonOption Text="Option A" Value="A"></BitRadioButtonOption>
    <BitRadioButtonOption Text="Option B" Value="B"></BitRadioButtonOption>
    <BitRadioButtonOption Text="Disabled option C" Value="C" IsEnabled="false"></BitRadioButtonOption>
</BitRadioButtonGroup>

@code {
    private string MyValue = "B";
}

```

Cyrus-Sushiant:
This closes #2541

### DIFF
--- a/src/BlazorUI/Bit.BlazorUI/Components/BitInputBase.cs
+++ b/src/BlazorUI/Bit.BlazorUI/Components/BitInputBase.cs
@@ -90,7 +90,7 @@ public abstract class BitInputBase<TValue> : BitComponentBase, IDisposable
         get => _value;
         set
         {
-            var hasChanged = !EqualityComparer<TValue>.Default.Equals(value, _value);
+            var hasChanged = EqualityComparer<TValue>.Default.Equals(value, _value) is false;
             if (hasChanged)
             {
                 _value = value;
--- a/src/BlazorUI/Bit.BlazorUI/Components/BitInputBase.cs
+++ b/src/BlazorUI/Bit.BlazorUI/Components/BitInputBase.cs
@@ -1,16 +1,12 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Diagnostics.CodeAnalysis;
-using System.Linq;
+﻿using System.Diagnostics.CodeAnalysis;
 using System.Linq.Expressions;
-using System.Threading.Tasks;
-using Microsoft.AspNetCore.Components;
 using Microsoft.AspNetCore.Components.Forms;
 
 namespace Bit.BlazorUI;
 
 public abstract class BitInputBase<TValue> : BitComponentBase, IDisposable
 {
+    private TValue? _value;
     private bool? _valueInvalid;
     private Type? _nullableUnderlyingType;
     private bool _hasInitializedParameters;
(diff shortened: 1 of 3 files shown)


src/BlazorUI/Bit.BlazorUI/Components/BitInputBase.cs
```diff
@@ -66,14 +87,14 @@ private set
 
     protected TValue? CurrentValue
     {
-        get => Value;
+        get => _value;
         set
         {
-            var hasChanged = !EqualityComparer<TValue>.Default.Equals(value, Value);
+            var hasChanged = !EqualityComparer<TValue>.Default.Equals(value, _value);
```
msynk, line 93: please use `is false` instead of `!` operator

src/BlazorUI/Bit.BlazorUI/Components/RadioButtonGroup/BitRadioButtonGroup.razor.cs
```diff
@@ -134,6 +154,38 @@ private void SelectOptionByKey(string? value)
 
     private string GetAriaLabelledBy() => Label.HasValue() || LabelFragment is not null ? LabelId : AriaLabelledBy;
 
+    private void HandleOnValueChanging(object? sender, ValueChangingEventArgs<string?> args)
+    {
+        if (firstRender) return;
```
msynk, line 159: what is this???

msynk (changes requested):
plz resolve the above comments

src/BlazorUI/Bit.BlazorUI/Components/RadioButtonGroup/BitRadioButtonGroup.razor.cs
Cyrus-Sushiant: The first time the Event is called because the value of the `AllOptions` is not yet initialized, the execution of this method should be skipped and this part should be managed by the OnAfterRenderAsync method.

msynk: you can just check the AllOptions then, no need to introduce another field for this class

msynk:
@Cyrus-Sushiant ???

src/BlazorUI/Bit.BlazorUI/Components/RadioButtonGroup/BitRadioButtonGroup.razor.cs
Cyrus-Sushiant: We do this by checking that the item variable `AllOptions` has an item, when the component does not have an item and the Value has a value, then the Value will not be reset.

msynk: so remove that redundant variable

msynk (approved):

PR #2676, opened 2022-07-22
Add placeholder unit test for NumericTextField #2244

### LINKED ISSUE #2244: Unit Tests for the new `Placeholder` parameter of the `BitNumericTextField` component
It is required to provide suitable unit tests for the `Placeholder` parameter of the `BitNumericTextField `component, which will be implemented in #2243.

Md23Mh:
This closes #2244


msynk (changes requested):
It seems you've copied a single test method to all types of the NumbericTextField tests.
it is better to have a general test for these types of tests, so we don't have to implement these tests in all of the other ones.

msynk:
@Md23Mh ?

Md23Mh:
Yep precisely! I did it because it is what apparently happened to some test methods for this component, and I just followed the rule here.

msynk:
@Md23Mh 
yeah, I know about other methods. but please update the PR based on my previous comment.

msynk (approved):

PR #2772, opened 2022-07-27
Fix the incorrect `async void` usage in the Bit Platform projects (#2765)

### LINKED ISSUE #2765: Incorrect `async void` usage in the Bit Platform projects
there are several places all over the Bit Platform projects that use `async void` which is a dangerous bad practice. this might result in unhandled and un-catchable exceptions. it is required to replace these `async void` with `async Task` ASAP.

sajadkardel:
this closes #2765


msynk (changes requested):

msynk:
@sajadkardel 
please use the appropriate `WrapHandled` method for calling these methods in the razor files in the template projects.

sajadkardel:
> @sajadkardel please also use the appropriate `WrapHandled` methods for calling the methods (that you've changed) in the razor files of the bit platform projects if not used.

It was one. in the `ProductsPage.razor` file and I fixed it. @msynk

msynk (approved):

ysmoradi (approved):

PR #2855, opened 2022-08-02
Fix the incorrect specificity in the `BitNav` component styles (#2854)

### LINKED ISSUE #2854: Incorrect `specificity` in the `BitNav` component styles
Currently, the specificity in the styles of the `BitNav` component is so low that any global styling can override it easily. it is required to improve this specificity a little to overcome this problem.

sajadkardel:
this closes #2854

### DIFF
--- a/src/BlazorUI/Playground/Bit.BlazorUI.Playground/Web/Pages/Components/Nav/BitNavDemo.razor
+++ b/src/BlazorUI/Playground/Bit.BlazorUI.Playground/Web/Pages/Components/Nav/BitNavDemo.razor
@@ -23,7 +23,7 @@
                    box-sizing: border-box;
                    border: 1px solid #eee;
                    overflow-y: auto;"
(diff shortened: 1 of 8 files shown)


src/BlazorUI/Playground/Bit.BlazorUI.Playground/Web/Pages/Components/Nav/BitNavDemo.razor
```diff
@@ -23,7 +23,7 @@
                    box-sizing: border-box;
                    border: 1px solid #eee;
                    overflow-y: auto;"
-                    Class="bit-nav-wrapped-link"
+                    Class="wrapped-link"
```
msynk, line 26: hmm, I don't think this change is correct since it's a new CSS class defined in the demo project and not one of the component's classes.

msynk (changes requested):

msynk (approved):

PR #2868, opened 2022-08-02
Fix the text that does not fit in the item box of the `BitNav` component (#2853)

### LINKED ISSUE #2853: The text of the nav item does not fit in the item box of the `BitNav` component
If the text is too long, it will go beyond the specified box and break.

Untitled

sajadkardel:
this closes #2853


msynk (changes requested):
I don't agree with this solution. it's different from what FluentUI has implemented.

sajadkardel:
I closed this PR, because the issue related to it, is not a valid issue.

PR #2872, opened 2022-08-03
Fix `BitButton` and friends to be type="submit" in a form via default

### LINKED ISSUE #2874: `BitButton` should trigger form onsubmit when inside a form
### Discussed in https://github.com/bitfoundation/bitplatform/discussions/2866

<div type='discussions-op-text'>

<sup>Originally posted by **linkdotnet** August  3, 2022</sup>
I am not sure if this is really feasible and I do understand that the idea is arguable at best. But anyway:

Any button, which doesn't define explicitly the `type="button"`, will be of `type="submit"`. 

> The missing value default and invalid value default are the Submit Button state.

Specification can be found here: https://html.spec.whatwg.org/multipage/form-elements.html#attr-button-type

That means if we have a component like this:
```razor
<EditForm Model="@MyModel" OnValidSubmit="HandleSubmit">
    <BitButton>Submit</BitButton>
</EditForm>

@code {
    private void HandleSubmit()
    {
        // Something to be invoked here
    }
}
```

Clicking on the button would invoke `HandleSubmit`. Only explicitly setting it to `ButtonType="BitButtonType.Submit"`. 
Now I do understand that the default is `ButtonType.Button` which translates to `type="button"`. The confusion comes from the point that if you would do that in html (without specifing the type) the form would be submitted.

One possible solution would be to make `ButtonType` optional aka nullable and check whether or not a `EditContext` is available.</div>

linkdotnet:
this closes #2874 

This PR addressed the issue, which was in the discussion #2866 .

I took the liberty and created a small PR, which might go in the right direction. Feedback very welcome.

### DIFF
--- a/src/BlazorUI/Bit.BlazorUI.Tests/Buttons/BitButtonTests.cs
+++ b/src/BlazorUI/Bit.BlazorUI.Tests/Buttons/BitButtonTests.cs
@@ -181,7 +181,7 @@ public class BitButtonTests : BunitTestContext
     [TestMethod]
     public void BitButtonSubmitStateInEditContextTest()
     {
-        var com = RenderComponent<BitButtonTest>(parameters =>
+        var com = RenderComponent<BitButton>(parameters =>
         {
             parameters.Add(p => p.EditContext, new EditContext(this));
         });
@@ -194,7 +194,7 @@ public class BitButtonTests : BunitTestContext
     [TestMethod]
     public void BitButtonButtonStateNotOverridenInEditContextTest()
     {
-        var com = RenderComponent<BitButtonTest>(parameters =>
+        var com = RenderComponent<BitButton>(parameters =>
         {
             parameters.Add(p => p.EditContext, new EditContext(this));
             parameters.Add(p => p.ButtonType, BitButtonType.Button);
--- a/src/BlazorUI/Bit.BlazorUI.Tests/Buttons/BitButtonTests.cs
+++ b/src/BlazorUI/Bit.BlazorUI.Tests/Buttons/BitButtonTests.cs
@@ -1,4 +1,5 @@
 ﻿using Bunit;
+using Microsoft.AspNetCore.Components.Forms;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
 namespace Bit.BlazorUI.Tests.Buttons;
@@ -156,7 +157,7 @@ public class BitButtonTests : BunitTestContext
 
         var bitButton = com.Find(".bit-btn");
 
-        Assert.AreEqual(bitButton.HasAttribute("aria-hidden"), expectedResult);
+        Assert.AreEqual(expectedResult, bitButton.HasAttribute("aria-hidden"));
     }
 
     [DataTestMethod,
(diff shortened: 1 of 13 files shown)


src/BlazorUI/Bit.BlazorUI.Tests/Buttons/BitButtonTests.cs
```diff
@@ -156,7 +157,7 @@ public void BitButtonAriaHiddenTest(bool ariaHidden, bool expectedResult)
 
         var bitButton = com.Find(".bit-btn");
 
-        Assert.AreEqual(bitButton.HasAttribute("aria-hidden"), expectedResult);
+        Assert.AreEqual(expectedResult, bitButton.HasAttribute("aria-hidden"));
```
linkdotnet, line 160: Normally expected comes first then actual.

I can also go through the tests and align this everywhere.

src/BlazorUI/Bit.BlazorUI.Tests/Buttons/BitButtonTests.cs
```diff
@@ -174,6 +175,33 @@ public void BitButtonTypeOfButtonTest(BitButtonType buttonType)
...
         var buttonTypeName = buttonType == BitButtonType.Button ? "button" : buttonType == BitButtonType.Submit ? "submit" : "reset";
-        Assert.AreEqual(bitButton.GetAttribute("type"), buttonTypeName);
+        Assert.AreEqual(buttonTypeName, bitButton.GetAttribute("type"));
     }
-}
+
+    [TestMethod]
+    public void BitButtonSubmitStateInEditContextTest()
+    {
+        var com = RenderComponent<BitButton>(parameters =>
```
linkdotnet, line 184: I explicitly did not test against `BitButtonTest` as directly testing is way easier.

In theory (in a follow up PR) we could remove this `BitButtonTest` and friends completely. The only reason is to have a `CascadingParameter` which we could add via `p => p.AddCascadingValue` anyway.

src/BlazorUI/Bit.BlazorUI.Tests/Buttons/BitButtonTests.cs
msynk: tnx, that would be great 🚀

src/BlazorUI/Bit.BlazorUI.Tests/Buttons/BitButtonTests.cs
msynk: yeah, you are right about these ...Test.razor component completely.
we kept adding these components as a good practice for future requirements (if any come up), but for now, we don't have any specific use for them.
although these middleware components are creating some issues/bugs testing 2-way parameters. let's discuss it in another thread.

src/BlazorUI/Bit.BlazorUI.Tests/Buttons/BitButtonTests.cs
linkdotnet: Said and done.

msynk (approved):
tnx a lot, great job 🚀🚀🚀

linkdotnet:
> tnx a lot, great job 🚀🚀🚀

Thanks as well for the super good support. I will leave it up to you guys whether or not you want to merge this. Or better when ;)

ysmoradi (approved):

msynk:
> > tnx a lot, great job 🚀🚀🚀
> 
> Thanks as well for the super good support. I will leave it up to you guys whether or not you want to merge this. Or better when ;)

oh ofc, we are going to merge it (squash it here😉)

Issue #2876, opened 2022-08-03
Redundant Test components for button components in the Bit BlazorUI Test project


msynk:
For each button component in the Test project, a Test component is added which currently there is no use for them. in order to have a simpler testing experience, it is better to remove them, and use the component directly in the unit test methods.

PR #2897, opened 2022-08-04
Feature/align tests

linkdotnet:
Another follow-up, which we discussed in #2872 / #2876:

This time I removed the further wrapper components and simplified the tests. I changed roughly 1/3 of the tests. Refactoring more would make the PR even bigger. I wanted to have it smaller packages.

EDIT: Please only consider this PR if you would anyway go into this direction and remove the wrapper components.


src/BlazorUI/Bit.BlazorUI.Tests/Checkboxes/BitCheckboxTests.cs
```diff
@@ -27,11 +27,15 @@ public class BitCheckboxTests : BunitTestContext
     ]
     public void BitCheckboxOnClickShouldWorkIfIsEnabled(Visual visual, bool defaultValue, bool isEnabled)
     {
-        var component = RenderComponent<BitCheckboxTest>(parameters =>
+        var clicked = false;
+        var changed = false;
+        var component = RenderComponent<BitCheckbox>(parameters =>
         {
             parameters.Add(p => p.Visual, visual);
```
Cyrus-Sushiant, line 34: I think you should use `AddCascadingValue` instead of `Add` to send data to the `Visual` parameter. The `Visual` is cascading parameter.

src/BlazorUI/Bit.BlazorUI.Tests/Checkboxes/BitCheckboxTests.cs
```diff
@@ -59,7 +60,7 @@ public void BitCheckboxOnClickShouldWorkIfIsEnabled(Visual visual, bool defaultV
     ]
     public void BitCheckboxBoxSideTest(Visual visual, BitCheckBoxSide boxSide)
     {
-        var component = RenderComponent<BitCheckboxTest>(parameters =>
+        var component = RenderComponent<BitCheckbox>(parameters =>
         {
             parameters.Add(p => p.Visual, visual);
```
Cyrus-Sushiant, line 65: Plz, Use AddCascadingValue instead of Add for the `Visual` parameter.

src/BlazorUI/Bit.BlazorUI.Tests/FileUpload/BitFileUploadTests.cs
```diff
@@ -14,7 +14,7 @@ public class BitFileUploadTests : BunitTestContext
     ]
     public void BitUploadFile_HasBasicClasses(Visual visual)
     {
-        var com = RenderComponent<BitFileUploadTest>(parameters =>
+        var com = RenderComponent<BitFileUpload>(parameters =>
         {
             parameters.Add(p => p.Visual, visual);
```
Cyrus-Sushiant, line 19: Plz, Use AddCascadingValue instead of Add for the `Visual` parameter.

src/BlazorUI/Bit.BlazorUI.Tests/Inputs/BitChoiceGroupTests.cs
```diff
@@ -25,14 +26,14 @@ public class BitChoiceGroupTests : BunitTestContext
    ]
     public void BitChoiceGroupShouldRespectIsEnabled(Visual visual, bool groupIsEnabled, bool optionIsEnabled)
     {
-        var component = RenderComponent<BitChoiceGroupTest>(
+        var component = RenderComponent<BitChoiceGroup>(
             parameters =>
             {
                 parameters.Add(p => p.Visual, visual);
```
Cyrus-Sushiant, line 32: Plz, Use AddCascadingValue instead of Add for the `Visual` parameter.

src/BlazorUI/Bit.BlazorUI.Tests/Inputs/BitDropDownTests.cs
```diff
@@ -25,7 +25,7 @@ public void BitDropDownShouldTakeCorrectVisual(Visual visual, bool isEnabled)
     {
         Context.JSInterop.Mode = JSRuntimeMode.Loose;
 
-        var component = RenderComponent<BitDropDownTest>(parameters =>
+        var component = RenderComponent<BitDropDown>(parameters =>
         {
             parameters.Add(p => p.Visual, visual);
```
Cyrus-Sushiant, line 30: Plz, Use AddCascadingValue instead of Add for the `Visual` parameter.

src/BlazorUI/Bit.BlazorUI.Tests/Inputs/BitDropDownTests.cs
```diff
@@ -53,7 +53,7 @@ public void ResponsiveDropDownShouldTakeCorrectClassName(Visual visual, bool isR
...
```
Cyrus-Sushiant, line 58: Plz, Use AddCascadingValue instead of Add for the `Visual` parameter.

src/BlazorUI/Bit.BlazorUI.Tests/Inputs/BitDropDownTests.cs
```diff
@@ -103,7 +104,7 @@ public void BitDropDownIsMultiSelectShouldWorkCorrect(Visual visual, bool isMult
         Context.JSInterop.Mode = JSRuntimeMode.Loose;
 
         var items = GetDropdownItems();
-        var component = RenderComponent<BitDropDownTest>(parameters =>
+        var component = RenderComponent<BitDropDown>(parameters =>
         {
             parameters.Add(p => p.IsOpen, true);
             parameters.Add(p => p.Visual, visual);
```
Cyrus-Sushiant, line 110: Plz, Use AddCascadingValue instead of Add for the `Visual` parameter.

src/BlazorUI/Bit.BlazorUI.Tests/Inputs/BitDropDownTests.cs
```diff
@@ -454,31 +455,23 @@ public void BitDropDownNotifyOnReselectShouldWorkCorrect(Visual visual, bool not
     {
         Context.JSInterop.Mode = JSRuntimeMode.Loose;
         var items = GetRawDropdownItems();
-        var component = RenderComponent<BitDropDownTest>(parameters =>
+        var itemSelecetd = false;
+        var component = RenderComponent<BitDropDown>(parameters =>
         {
             parameters.Add(p => p.Visual, visual);
```
Cyrus-Sushiant, line 461: Plz, Use AddCascadingValue instead of Add for the `Visual` parameter.

Cyrus-Sushiant (changes requested):

src/BlazorUI/Bit.BlazorUI.Tests/Checkboxes/BitCheckboxTests.cs
linkdotnet: Actually it doesn't matter in that case and `Add` will call `AddCascading` see [here](https://github.com/bUnit-dev/bUnit/blob/main/src/bunit.core/ComponentParameterCollectionBuilder.cs#L49). bUnit does detect if it is a cascading one or not.

See it like that: From your component under test it does not matter, as we set a property. If the property has attribute A or B we don't care.
Still the question is: Why do we care then? We still have that switch, plus we have `AddCascadingValue`. The reason is simple:
```razor
<CascadingValue ...>
    <ComponentUnderTest>
        <ChildComponent>
        </ChildComponent>
    </ComponentUnderTest>
</CascadingValue>
```

Now if `ChildComponent` uses the cascading parameter but `ComponentUnderTest` we need a way to create those. And that is why you have `AddCascadingValue`.

To recap: If your component uses the cascading value directly and you have no sibling relying on that...you can use directly `Add`. If a sibling needs the parameter, then you have to use `AddCascadingValue`.

So in this case, it is preference. Sure I can change to `AddCascadingValue` if you want. Your choice ;)

msynk: yeah, you're right. from the component pov it doesn't matter that the parameter is coming from what source.
but for the sake of compatibility/consistency and clarity, let's keep the cascading parameters being set by the `AddCascadingValue` method.

linkdotnet: I changed it to `AddCascadingValue`. Predominantly you use `Add` instead of `AddCascadingValue` in your tests. I could align that in a follow up if wished.

Cyrus-Sushiant (approved):

msynk (approved):

PR #2908, opened 2022-08-06
show foribidden error message when user do something without desired permissions (#2907)

### LINKED ISSUE #2907: Incorrect navigation to the sign-in page when the current user wants to navigate a page without sufficient permissions in the template projects

ysmoradi:
This closes #2907


src/Templates/AdminPanel/Bit.AdminPanel/Web/App.razor
```diff
@@ -5,9 +5,9 @@
                 AdditionalAssemblies="@lazyLoadedAssemblies"
                 OnNavigateAsync="@OnNavigateAsync">
             <Found Context="routeData">
-                <AuthorizeRouteView RouteData="@routeData">
+                <AuthorizeRouteView Context="authState" RouteData="@routeData">
```
msynk, line 8: please use the AuthenticationState cascading parameter instead of this parameter passing.

msynk (dismissed):

msynk (approved):

PR #4050, opened 2023-03-07
Fix MacCatalyst application title (#4015)

### LINKED ISSUE #4015: `Mac Catalyst` Title is not correct
The app doesn't show up in macOS's launchpad, but you can see it in applications folder, because the app name is not correct; it shows csproject name instead (TodoTmplate.Client.App) - [MAUI Issues](https://github.com/dotnet/maui/issues/12287)

mhrastegari:
closes #4015


ysmoradi (changes requested):
I'm pretty sure that AssemblyName is not what we're looking for

msynk:
we need to talk about it

ysmoradi (approved):

msynk (approved):

PR #4468, opened 2023-05-28
Fix iOS zooming on inputs in web projects (#4467)

### LINKED ISSUE #4467: Browser zooms on Inputs in iOS
In Web versions, `iOS` browsers getting focus on Input components that their font sizes are less than 16px.

mhrastegari:
Closes #4467


mhrastegari:
After some tests it's working fine on iOS and Windows browsers but in Android zooming is fully locked!!! 
So, I think it's risky and better to closing this PR.

PR #5255, opened 2023-09-06
Suppress null warning

### LINKED ISSUE #5224: Possible null exception
The value of [restError](https://github.com/bitfoundation/bitplatform/blob/af98af8c20d47d73d49109d848be365f6fa76eef/src/Websites/Careers/src/Bit.Websites.Careers.Web/Services/Implementations/AppHttpClientHandler.cs#L28C26-L28C26) may sometimes be `null`, although it is not expected. However, [ReadFromJsonAsync](https://github.com/bitfoundation/bitplatform/blob/af98af8c20d47d73d49109d848be365f6fa76eef/src/Websites/Careers/src/Bit.Websites.Careers.Web/Services/Implementations/AppHttpClientHandler.cs#L28C26-L28C26) could still return `null`. Therefore, we need to ensure that the values of `restError.Key` and `restError.Message` are not null as well.

Although it is unlikely that we will receive `null`, we can fix this issue by either suppressing null with `!` (_fastest, but not safest!_) or double-checking all possible null properties to ensure they are not null (_suggested one although it could be unnecessary!_).

iPazooki:
this closes #5224 

### Suppress null warning as we are sure middleware will not return null

This pull request will suppress null warnings for `AppHttpClientHandler.cs` as we discussed in this issue: #5224

### Checklist before requesting a review

- [x] I have performed a self-review of my code

### Changes

- Use suppress null operator `!` to suppress null warnings


msynk (approved):

msynk:
did you check the same class in other projects?
I think we might need to apply the same changes for some of them.

@ysmoradi I've checked these classes and found the implementation is different for some of them.
is it intentional?

iPazooki:
Aha, I was not aware of that. I'll deal with it

PR #5316, opened 2023-09-20
Add BitSnackBar Tests (#4402)

### LINKED ISSUE #4402: Add `BitSnackbar` Tests
Add `BitSnackbar` Tests

0xSamman:
I hope this closes #4402

### DIFF
--- a/src/BlazorUI/Bit.BlazorUI.Tests/SnackBar/BitSnackBarTests.cs
+++ b/src/BlazorUI/Bit.BlazorUI.Tests/SnackBar/BitSnackBarTests.cs
@@ -1,0 +1,205 @@
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using Bunit;
+using Bunit.Extensions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+namespace Bit.BlazorUI.Tests.SnackBar;
+
+[TestClass]
+public class BitSnackBarTests : BunitTestContext
+{
+    [TestMethod]
+    public void BitSnackBarInitialTest()
+    {
+        var com = RenderComponent<BitSnackBar>();
+
+        var element = com.Find(".bit-snb");
+
+        Assert.IsNotNull(element);
+    }
+
+    [DataTestMethod,
+        DataRow(BitSnackBarPosition.TopLeft),
+        DataRow(BitSnackBarPosition.TopCenter),
+        DataRow(BitSnackBarPosition.TopRight),
+        DataRow(BitSnackBarPosition.BottomLeft),
+        DataRow(BitSnackBarPosition.BottomCenter),
+        DataRow(BitSnackBarPosition.BottomRight),
+        DataRow(null)
+    ]
+    [TestMethod]
+    public void BitSnackBarPositionTest(BitSnackBarPosition? position)
+    {
+        var com = RenderComponent<BitSnackBar>(parameters =>
+        {
+            if (position.HasValue) parameters.Add(p => p.Position, position.Value);
+        });
+
+        var element = com.Find(".bit-snb");
+
+        var positionClass = position switch
+        {
+            BitSnackBarPosition.TopLeft => "bit-snb-tlf",
+            BitSnackBarPosition.TopCenter => "bit-snb-tcn",
+            BitSnackBarPosition.TopRight => "bit-snb-trt",
+
+            BitSnackBarPosition.BottomLeft => "bit-snb-blf",
+            BitSnackBarPosition.BottomCenter => "bit-snb-bcn",
+            BitSnackBarPosition.BottomRight => "bit-snb-brt",
+
+            _ => "bit-snb-brt",
+        };
+
+        Assert.IsTrue(element.ClassList.Contains(positionClass));
+    }
+
+    [DataTestMethod,
+        DataRow("title", "body"),
+        DataRow("title", "")
+    ]
+    [TestMethod]
+    public async Task BitSnackBarShowTest(string title, string body)
+    {
+        var com = RenderComponent<BitSnackBar>();
+
+        await com.Instance.Show(title, body);
+
+        var item = com.Find(".bit-snb-itm");
+        Assert.IsNotNull(item);
+
+        var titleElement = com.Find(".bit-snb-ttl");
+        Assert.AreEqual(title, titleElement.InnerHtml);
+
+        if (body.IsNullOrEmpty() is false)
+        {
+            var bodyElement = com.Find(".bit-snb-bdy");
+            Assert.AreEqual(body, bodyElement.InnerHtml);
+        }
+    }
+
+    [DataTestMethod,
+        DataRow(true),
+        DataRow(false)
+    ]
+    [TestMethod]
+    public async Task BitSnackBarAutoDismissTest(bool autoDismiss)
+    {
+        var com = RenderComponent<BitSnackBar>(
+            parameters =>
+            {
+                parameters.Add(p => p.AutoDismiss, autoDismiss);
+            }
+        );
+
+        await com.Instance.Show("title");
+
+        //Added a sec delay to be sure we are asserting after AutoDismissTime passed
+        await Task.Delay(com.Instance.AutoDismissTime + TimeSpan.FromSeconds(1));
+
+        var items = com.FindAll(".bit-snb-itm");
+
+        Assert.AreEqual(autoDismiss ? 0 : 1, items.Count);
+    }
+
+    [DataTestMethod,
+        DataRow(BitSnackBarType.Info),
+        DataRow(BitSnackBarType.Warning),
+        DataRow(BitSnackBarType.Success),
+        DataRow(BitSnackBarType.Error),
+        DataRow(BitSnackBarType.SevereWarning),
+        DataRow(null)
+    ]
+    [TestMethod]
+    public async Task BitSnackBarTypeTest(BitSnackBarType? type)
+    {
+        var com = RenderComponent<BitSnackBar>();
+
+        if (type.HasValue)
+            await com.Instance.Show("title", string.Empty, type);
+        else
+            await com.Instance.Show("title");
+
+        var element = com.Find(".bit-snb-itm");
+
+        var typeClass = type switch
+        {
+            BitSnackBarType.Info => $"bit-snb-info",
+            BitSnackBarType.Warning => $"bit-snb-warning",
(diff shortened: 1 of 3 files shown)


src/BlazorUI/Bit.BlazorUI.Tests/SnackBar/BitSnackBarTests.cs
```diff
@@ -0,0 +1,205 @@
...
+    ]
+    [TestMethod]
+    public async Task BitSnackBarTypeTest(BitSnackBarType? type)
+    {
+        var com = RenderComponent<BitSnackBar>();
+
+        if (type.HasValue)
+            await com.Instance.Show("title", string.Empty, type);
+        else
+            await com.Instance.Show("title");
```
msynk, line 121: plz always use `{` and `}` for blocks.
there are two exceptions for this rule of one-liners: `return` & `throw`

src/BlazorUI/Bit.BlazorUI.Tests/SnackBar/BitSnackBarTests.cs
```diff
@@ -0,0 +1,205 @@
...
+        Assert.AreEqual(0, items.Count);
+    }
+
+    [TestMethod]
+    public async Task BitSnackBarDismissIconNameTest()
+    {
+        var com = RenderComponent<BitSnackBar>(
+            parameters =>
+            {
+                parameters.Add(x => x.DismissIconName, "Go");
```
msynk, line 159: about this string `Go` here:
always introduce these expected values as method arguments (test data), and plz do not use repeating string values like this.

src/BlazorUI/Bit.BlazorUI.Tests/SnackBar/BitSnackBarTests.cs
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    [TestMethod]
+    public async Task BitSnackBarTitleTemplateTest()
+    {
+        var wrappedCom = RenderComponent<BitSnackBarWithTitleTemplate>();
+
+        var com = wrappedCom.FindComponent<BitSnackBar>();
+
+        await com.Instance.Show("title");
```
msynk, line 177: again, plz change this expected value into the test method input argument.
also change that `second element` string into a parameter like Title, since it is also an expected value.

src/BlazorUI/Bit.BlazorUI.Tests/SnackBar/BitSnackBarTests.cs
```diff
@@ -0,0 +1,205 @@
...
+
+        var titleTemplateElements = com.FindAll(".bit-snb-hdr > *").ToList();
+        //remove button
+        titleTemplateElements.RemoveAt(0);
+
+        Assert.AreEqual("<span>title</span><span>second element</span>", string.Join(string.Empty, titleTemplateElements.Select(x => x.OuterHtml).ToArray()));
```
msynk, line 183: why not use the InnerHtml of the parent instead of combining the OuterHtml of children?

src/BlazorUI/Bit.BlazorUI.Tests/SnackBar/BitSnackBarTests.cs
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    [TestMethod]
+    public async Task BitSnackBarBodyTemplateTest()
+    {
+        var wrappedCom = RenderComponent<BitSnackBarWithBodyTemplate>();
+
+        var com = wrappedCom.FindComponent<BitSnackBar>();
+
+        await com.Instance.Show("title", "body");
```
msynk, line 193: again, expected values and method argument.

src/BlazorUI/Bit.BlazorUI.Tests/SnackBar/BitSnackBarTests.cs
```diff
@@ -0,0 +1,205 @@
...
+
+        var itemTemplateElements = com.FindAll(".bit-snb-itm > *").ToList();
+        //remove title
+        itemTemplateElements.RemoveAt(0);
+        //remove progressbar
+        itemTemplateElements.RemoveAt(itemTemplateElements.Count - 1);
+
+        Assert.AreEqual("<p>body</p><span>second element</span>", string.Join(string.Empty, itemTemplateElements.Select(x => x.OuterHtml).ToArray()));
```
msynk, line 201: again, parent InnerHtml instead of children's OuterHtml.

src/BlazorUI/Bit.BlazorUI.Tests/SnackBar/BitSnackBarWithBodyTemplate.razor
```diff
@@ -0,0 +1,6 @@
+<BitSnackBar>
+    <BodyTemplate Context="body">
+        <p>@body</p>
+        <span>second element</span>
```
msynk, line 4: either remove this line or accept it as a parameter, since it is part of the expected value in the test.

src/BlazorUI/Bit.BlazorUI.Tests/SnackBar/BitSnackBarWithTitleTemplate.razor
```diff
@@ -0,0 +1,6 @@
+<BitSnackBar>
+    <TitleTemplate Context="title">
+        <span>@title</span>
+        <span>second element</span>
```
msynk, line 4: again, either remove this line or accept it as a parameter, since it is part of the expected value in the test.

msynk (changes requested):

src/BlazorUI/Bit.BlazorUI.Tests/SnackBar/BitSnackBarTests.cs
0xSamman: The problem here is we have a Button here that is in the same level of Template content.
We should able to get those, I tried to make it possible with correct selector but I couldn't do that, so I removed it like that.

src/BlazorUI/Bit.BlazorUI.Tests/SnackBar/BitSnackBarTests.cs
0xSamman: same as the above problem, here we have an progressbar in additional

src/BlazorUI/Bit.BlazorUI.Tests/SnackBar/BitSnackBarWithBodyTemplate.razor
0xSamman: The idea behind this is when I tried to assert without removing unwanted children I have to select second child element to prevent Title header and progressbar. this will pass but when I was thinking what if user have two element which not wrapped to a element. So my test failed after adding this line and I've had to change the test to what you mentioned as a problem to removing child like this

0xSamman: The idea behind this is when I tried to assert without removing unwanted children I have to select second child element to prevent Title header and progressbar. this will pass but when I was thinking what if user have two element which not wrapped to a element. So my test failed after adding this line and I've had to change the test to what you mentioned as a problem to removing child like this

src/BlazorUI/Bit.BlazorUI.Tests/SnackBar/BitSnackBarTests.cs
0xSamman: sure, got it

src/BlazorUI/Bit.BlazorUI.Tests/SnackBar/BitSnackBarTests.cs
0xSamman: Of course, my bad

src/BlazorUI/Bit.BlazorUI.Tests/SnackBar/BitSnackBarTests.cs
msynk: check the bUnit docs especially this page to use a better approach for these tests:

https://bunit.dev/docs/verification/verify-component-state.html

src/BlazorUI/Bit.BlazorUI.Tests/SnackBar/BitSnackBarWithBodyTemplate.razor
```diff
@@ -0,0 +1,6 @@
+<BitSnackBar>
+    <BodyTemplate Context="body">
+        <p>@body</p>
+        <span></span>
```
msynk, line 4: plz, either remove this line or use it as a meaningful section of the test

src/BlazorUI/Bit.BlazorUI.Tests/SnackBar/BitSnackBarWithTitleTemplate.razor
```diff
@@ -0,0 +1,6 @@
+<BitSnackBar>
+    <TitleTemplate Context="title">
+        <span>@title</span>
+        <span></span>
```
msynk, line 4: again, either remove this line or use it as a meaningful section of the test

msynk (changes requested):

src/BlazorUI/Bit.BlazorUI.Tests/SnackBar/BitSnackBarTests.cs
msynk: again:

https://bunit.dev/docs/verification/verify-component-state.html
https://bunit.dev/docs/verification/semantic-html-comparison.html

src/BlazorUI/Bit.BlazorUI.Tests/SnackBar/BitSnackBarWithBodyTemplate.razor
msynk: https://bunit.dev/docs/verification/verify-component-state.html
https://bunit.dev/docs/verification/semantic-html-comparison.html

src/BlazorUI/Bit.BlazorUI.Tests/SnackBar/BitSnackBarTests.cs
0xSamman: Thanks a lot, I'll revise my code after read this

msynk:
I'm closing this PR since there was no activity for a long time.

PR #5495, opened 2023-10-08
Add appropriate code Highlighting in light mode #5472

### LINKED ISSUE #5472: Code Highlighting for samples in light mode
Currently, in the light mode, we use the dark color theme for the sample code which is not appropriate:

image

taha-ghadirian:
closes #5472 

Steps produced: 
1. Created a **scss** file in `styles/prism-1.28.0/prism-okaidia-bit.scss`
2. Added color variables for those colors who are appliable. 
3. Added `:root[bit-theme=...]`  check  for text shadow. 
4. Changed some hard coded colors, so they can match with both dark and white theme.


msynk:
Thanks for submitting this issue. but this approach is not quite correct for our project.
you shouldn't introduce scss files in the wwwroot folder. the only file that needs to be changed is the CSS file (that you deleted in this PR), and use the CSS variables instead of the hardcoded colors in it.

msynk (changes requested):
invalid approach

PR #5585, opened 2023-10-15
Fix incorrect date format in properties of JWTs (#5256)

### LINKED ISSUE #5256: Wrong date format in the ExpireDate of the JWT token
In the api, in the input of the token creation method, the ExpireDate argument, according to Visual Studio, should be UTC, but you have passed DateTime.Now.  Why?

image

msynk:
this closes #5256


ysmoradi (approved):

ysmoradi (changes requested):
Rejected! Use dotnet 8 time abstractions instead.

ysmoradi (approved):

msynk:
Let's implement the new .NET 8 approach in our new clean architect template in the near future.

PR #5691, opened 2023-10-19
Incorrect page title of the Sticky demo page

### LINKED ISSUE #5686: Incorrect page title of the `Sticky` demo page
Currently, the the title of the `Sticky` demo page starts with `BitSticky` which is not correct. The correct value does not have `Bit` at the start and should be `Sticky`.

https://github.com/bitfoundation/bitplatform/blob/4321ff327a51f2238e6114af139eb5965c4be35a/src/BlazorUI/Demo/Client/Core/Pages/Components/Sticky/BitStickyDemo.razor#L4

KriteshTimsina:
fixes #5686


msynk:
Thanks for submitting the PR, but the issues labeled with `good first issue` are only for the first contributors, and more than one PR for them from a single contributor is not acceptable for our repository.
I'm sorry but I have to reject and close this PR.

Issue #5400, opened 2023-09-29
Convert the type of the `Height` and `Width` parameters to string in the `BitImage` component


msynk:
Convert the type of the `Height` and `Width` parameters to string in the `BitImage` component and update the demo page accordingly:

![image](https://github.com/bitfoundation/bitplatform/assets/6169846/81de8a93-25c0-4144-85b5-e56f144f5e3c)

![image](https://github.com/bitfoundation/bitplatform/assets/6169846/ecaac53d-4b78-4ea5-bf47-e32cd5b9082d)

PR #5707, opened 2023-10-20
convert types BitImage #5400

RamsinChaabian:
Change type values #5400


msynk:
@RamsinChaabian Thanks for submitting this PR, but this PR is not for the first contribution or ppl who are not familiar with our conventions. I suggest starting with `good first issues` as mentioned in our README.

PR #5712, opened 2023-10-20
Fix incorrect page title of the Scrollable demo page (#5683)

### LINKED ISSUE #5683: Incorrect page title of the `Scrollable` demo page
Currently, the the title of the `Scrollable` demo page starts with `BitScrollable` which is not correct. The correct value does not have `Bit` at the start and should be `Scrollable`.

https://github.com/bitfoundation/bitplatform/blob/4321ff327a51f2238e6114af139eb5965c4be35a/src/BlazorUI/Demo/Client/Core/Pages/Components/ScrollablePane/BitScrollablePaneDemo.razor#L4

hanieh1995:
this closes #5683


msynk:
@hanieh1995 Thanks for submitting this PR, but the `good first issues` are only for the first contribution. So let new contributors take care of them. You can try resolving other issues, especially the `up for grabs` ones.

Issue #5674, opened 2023-10-18
Update the title of the `BitModal with position` section of the `Modal` demo page


msynk:
The current title of this section is too long and not appropriate:

![image](https://github.com/bitfoundation/bitplatform/assets/6169846/671b9613-8a2b-49b7-9dec-c3f9e7f30015)

This title should be changed to `Position`.

The demo file name is `BitModalDemo.razor` in the `Bit.BlazorUI.Demo.Client.Core` project.

PR #5726, opened 2023-10-21
Update the title #5674

mahditahanian:


msynk:
@mahditahanian Thanks for the PR, but it seems you've tried to resolve more than one `good first issue` in this PR which is not allowed here, these issues are for the first contributions, so let other new contributors participate to resolve them.
Please file a new PR for only one of these issues and please make sure your PR has enough information about the issue (in the title and also in the description).

PR #5751, opened 2023-10-22
1 add login page(#1)

### LINKED ISSUE #1: Non key property ("Id") in Dto problem
I've got a Dto that has a property named "Id" but it's not a Key and I'm gonna update it's value later using jaydata, but unfortunately it's way seems make some un handled problems

jacksonalexg99:
this closes #1


jacksonalexg99:
why ??

jacksonalexg99 (commented):
review..

PR #5781, opened 2023-10-23
Using DateTimeOffset instead of DateTime to instantiate DateTimeOffset (#5786)

### LINKED ISSUE #5786: Using DateTime to create with DateTimeOffset

Actually It's not a ``bug`` but the ``warning`` that dotnet shows suggesting that you replace ``DateTime`` with ``DateTimeOffset`` to better handle date and time values, particularly when working with different time zones or when precision is needed regarding the exact offset from Coordinated Universal Time (UTC).

Your code is already in the right direction since you're creating a ``DateTimeOffset`` object. However, directly using ``DateTime`` to create a ``DateTimeOffset`` object isn't wrong per se, but there are better ways to do it.

[BitDatePickerDemo.razor](https://github.com/bitfoundation/bitplatform/blob/develop/src/BlazorUI/Demo/Client/Core/Pages/Components/Inputs/DatePicker/BitDatePickerDemo.razor)
[BitDatePickerDemo.razor.cs](https://github.com/bitfoundation/bitplatform/blob/develop/src/BlazorUI/Demo/Client/Core/Pages/Components/Inputs/DatePicker/BitDatePickerDemo.razor.cs)
[BitDateRangePickerDemo.razor](https://github.com/bitfoundation/bitplatform/blob/develop/src/BlazorUI/Demo/Client/Core/Pages/Components/Inputs/DateRangePicker/BitDateRangePickerDemo.razor)
[BitDateRangePickerDemo.razor.cs](https://github.com/bitfoundation/bitplatform/blob/develop/src/BlazorUI/Demo/Client/Core/Pages/Components/Inputs/DateRangePicker/BitDateRangePickerDemo.razor.cs)

zandiarash:
this closes #5786

### DIFF
--- a/src/BlazorUI/Demo/Client/Core/Pages/Components/Inputs/DatePicker/BitDatePickerDemo.razor
+++ b/src/BlazorUI/Demo/Client/Core/Pages/Components/Inputs/DatePicker/BitDatePickerDemo.razor
@@ -309,7 +309,7 @@
 </ComponentDemo>
 
 @code {
-    private DateTimeOffset? selectedDate = new DateTimeOffset(new DateTime(2020, 1, 17), DateTimeOffset.Now.Offset);
+    private DateTimeOffset? selectedDate = new DateTimeOffset(2020, 1, 17, 0, 0, 0, DateTimeOffset.Now.Offset);
     private DateTimeOffset? selectedDateTime = DateTimeOffset.Now;
     private FormValidationDatePickerModel formValidationDatePickerModel = new();
     private string SuccessMessage = string.Empty;
--- a/src/BlazorUI/Demo/Client/Core/Pages/Components/Inputs/DatePicker/BitDatePickerDemo.razor.cs
+++ b/src/BlazorUI/Demo/Client/Core/Pages/Components/Inputs/DatePicker/BitDatePickerDemo.razor.cs
@@ -492,7 +492,7 @@ private async Task OpenCallout()
                Placeholder=""Select a date..."" />
 <BitLabel>Selected date: @selectedDate.ToString()</BitLabel>";
     private readonly string example9CsharpCode = @"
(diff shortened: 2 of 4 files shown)


zandiarash:
Actually It's not a ``bug`` but the ``warning`` that dotnet shows suggesting that you replace ``DateTime`` with ``DateTimeOffset`` to better handle date and time values, particularly when working with different time zones or when precision is needed regarding the exact offset from Coordinated Universal Time (UTC).

Your code is already in the right direction since you're creating a ``DateTimeOffset`` object. However, directly using ``DateTime`` to create a ``DateTimeOffset`` object isn't wrong per se, but there are better ways to do it.

msynk:
@zandiarash Thanks a lot for submitting this PR, but we do not file a PR without first discussing the problem in a separate issue. Basically, the PR is not the place to investigate the issue for us. So if you can please submit the issue for this problem too.
btw, we are already aware of the problem and have plans to fix it, but we very welcome external contributions like this too.

zandiarash:
> @zandiarash Thanks a lot for submitting this PR, but we do not file a PR without first discussing the problem in a separate issue. Basically, the PR is not the place to investigate the issue for us. So if you can please submit the issue for this problem too. btw, we are already aware of the problem and have plans to fix it (we have a few related issues/PRs), but we very welcome external contributions like this too.

Of course!
Initially, I wished to spare you the effort of reading the issue about a suggestion which isn't truly a bug.
Thank you for your guidance and clarity on the procedure. I understand the importance of discussing the issue in a separate forum before submitting a PR. I'll ensure to raise an issue detailing the problem shortly.
I'm glad to hear that the team is already working on it, and I'm always eager to contribute where I can. Let me know if there's any other way I can assist or align my contributions better with the team's strategies.

msynk:
@zandiarash Wow, great job, thanks a lot 🚀

msynk (approved):

Issue #5433, opened 2023-10-01
Incorrect title for the binding sample of the `BitToggleButton` demo page


msynk:
Incorrect title for the binding sample of the `BitToggleButton` demo page: 
`BitToggleButtonDemo.razor` file in the `Bit.BlazorUI.Demo.Client.Core` project.

![image](https://github.com/bitfoundation/bitplatform/assets/6169846/2640e860-8adf-4124-8299-8a62347ff341)

Currently, the title is `Controlled` which is not correct and should be `Binding`.

msynk:
not relevant anymore!

PR #5782, opened 2023-10-23
fix(blazor-ui): update toggle button demo title

Hossein-Mirazimi:
This Closes: #5433


msynk:
@Hossein-Mirazimi Thanks for submitting this PR, but the `good first issues` are only for the first contribution, and resolving more than one of them is not allowed here. So please let the other new contributors take care of the rest. 

PR #5785, opened 2023-10-23
Add missing Linux to comment in IServiceCollectionExtensions in Shared project of bit BlazorUI demo (#5641)

### LINKED ISSUE #5641: Missing `Linux` from the comment in the `IServiceCollectionExtensions` class in the `Shared` project of the `bit BlazorUI` demo project
There is a comment on line 9 of the `IServiceCollectionExtensions.cs` file in the `Shared` project of the bit BlazorUI project template that does not include `Linux`:

image

the correct state of the change should be something like this:
image

here is a link to this line on GitHub:
https://github.com/bitfoundation/bitplatform/blob/fcb83f2c6da2a1b42c792497b39f977426e96917/src/BlazorUI/Demo/Shared/Extensions/IServiceCollectionExtensions.cs#L9

ShamimShahraeini:
this closes #5641

### DIFF
--- a/src/BlazorUI/Demo/Shared/Extensions/IServiceCollectionExtensions.cs
+++ b/src/BlazorUI/Demo/Shared/Extensions/IServiceCollectionExtensions.cs
@@ -6,7 +6,7 @@ public static class IServiceCollectionExtensions
 {
     public static void AddSharedServices(this IServiceCollection services)
     {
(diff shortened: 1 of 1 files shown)


ShamimShahraeini:
here is the related issue:

https://github.com/bitfoundation/bitplatform/issues/5641

msynk (changes requested):
The change is not exactly what the issue wants

msynk (approved):

PR #5799, opened 2023-10-24
Change the `ToLower` method to `ToLowerInvariant` in the AdminPanel

### LINKED ISSUE #5766: Change the `ToLower` method to `ToLowerInvariant` in the AdminPanel project template
It's better to use the `ToLowerInvariant` method in this line:
https://github.com/bitfoundation/bitplatform/blob/41c76cb3addbb96f256854a0925fdae529da5716/src/Templates/AdminPanel/Bit.AdminPanel/src/Client/Web/Extensions/HttpRequestExtensions.cs#L26

alireza-dhp:
This closes #5766


msynk:
@alireza-dhp Thanks for submitting this PR, but resolving more than one issue in a single PR is not allowed in this repo. So if you can, please push a new PR to resolve only one issue.

Issue #5656, opened 2023-10-18
Update the title of the `Change tabs from outside of the component` section of the `Pivot` demo page


msynk:
The current title of this section is too long and not appropriate:

![image](https://github.com/bitfoundation/bitplatform/assets/6169846/8c882b2b-fb97-4a28-b05f-a059e6df6246)

This title should be changed to `Binding`.

The demo file name is `BitPivotDemo.razor` in the `Bit.BlazorUI.Demo.Client.Core` project.

msynk:
not relevant anymore!

PR #5816, opened 2023-10-25
fix(blazorUI): Update the title of the Change tabs from outside of th…

Poorya-Rajabi:
…e component section of the Pivot demo page #5656


msynk:
Thanks for submitting this PR, but the `good first issues` are only for the first contribution, and resolving more than one of them is not allowed here. So please let the other new contributors take care of the rest. 

PR #5841, opened 2023-10-27
Add missing Linux to comment in IServiceCollectionExtensions class in Core project of TodoTemplate (#5636)

### LINKED ISSUE #5636: Missing `Linux` from the comment in the `IServiceCollectionExtensions` class in the `Core` project of the `TodoTemplate` project template
There is a comment on line 9 of the `IServiceCollectionExtensions.cs` file in the `Core` project of the TodoTemplate project template that does not include `Linux`:

image

the correct state of the change should be something like this:
image

here is a link to this line on GitHub:
https://github.com/bitfoundation/bitplatform/blob/fcb83f2c6da2a1b42c792497b39f977426e96917/src/Templates/TodoTemplate/Bit.TodoTemplate/src/Client/Core/Extensions/IServiceCollectionExtensions.cs#L8

alireza-dhp:
This closes #5636


msynk:
@alireza-dhp Thanks for submitting this PR, but the change you applied is not exactly what the issue wants. please double-check the issue's comment and try again.

msynk (approved):

PR #5886, opened 2023-10-30
Resolve missing Linux from the comment in the IServiceCollectionExtensions class in the Core project of the AdminPanel project template (#5635)

reza-dehghani:
Hi, This closes (#5635 ).


msynk:
@reza-dehghani Thanks for submitting this PR, but the `good first issues` are only for the first contribution, and resolving more than one of them is not allowed here. So please let the other new contributors take care of the rest. 

PR #5908, opened 2023-10-31
Update README.md

nobunagaaa:
In the "How to contribute?" section, the section header is underlined with asterisks, which might not render as expected in all environments. You can use a more common Markdown format for section headers, such as a single or double hash (e.g., ## or ###).


msynk (commented):
I'm sorry, but this change is not accepted.

PR #5980, opened 2023-11-08
Fix incorrect page title of the Sticky demo (#5686)

### LINKED ISSUE #5686: Incorrect page title of the `Sticky` demo page

Salamesalehpour:
this closes #5686


msynk (approved):

msynk:
@Salamesalehpour Thanks for your contribution 🚀

Issue #5653, opened 2023-10-18
Update the title of the `Controlled BitBreadcrumb` section of the `Breadcrumb` demo page


msynk:
The current title of this section is too long and not appropriate:

![image](https://github.com/bitfoundation/bitplatform/assets/6169846/c4c54b46-f3b8-46c5-91a7-d3cb4b289fb5)

This title should be changed to `Events`.

The demo file name is `BitBreadcrumbDemo.razor` in the `Bit.BlazorUI.Demo.Client.Core` project.

**Note**: There are three different locations for this change on this demo page.

PR #5990, opened 2023-11-10
update the title of the controlled bit bread crumb section of bread crumb demo page (#5653)

progman246:
the title of the controlled bit bread crumb section of bread crumb demo page updated.


msynk:
@progman246 Thanks for submitting this PR, but the `good first issues` are only for the first contribution, and resolving more than one of them is not allowed here. So please let the other new contributors take care of the rest. 

PR #7372, opened 2024-04-13
Add missing System.Net.Http.Json using in the generated code of HttpClient source generator (#7420)

### LINKED ISSUE #7420: Http client source generator relies on `System.Net.Http.Json` but doesn't explicitly uses that namespace
bit source generator relies on the following namespaces:
```
System.Text.Json
System.Net.Http.Json
System.Web
```
In consumer's project has the same namespaces as its own global usings, the generated code will be built, but if not, the project's build will not be successful. 
It would be a good idea to use these namespaces in generated code explicitly.

jpmchia:
closes #7420


msynk:
@jpmchia Thanks for submitting this PR.
but we do not accept PRs without discussing the requested changes in an issue first.

ysmoradi (approved):

ysmoradi:
This closes #7420

msynk (approved):

msynk:
after some investigation by our team, we discovered the problem and filed a new issue for it: #7420 
@jpmchia Thanks for the PR, it is accepted now.

PR #7403, opened 2024-04-24
Resolve animation direction issue in BitProgressIndicator (#7402)

### LINKED ISSUE #7402: Incorrect animation direction in `BitProgressIndicator` in RTL mode
When using the `BitProgressIndicator` component in Indeterminate mode with right-to-left (RTL) text direction, the animation behaves unexpectedly. Instead of moving from right to left (as expected in RTL mode), it currently moves from left to right, which is incorrect.

Cyrus-Sushiant:
This closes #7402

### DIFF
--- a/src/BlazorUI/Bit.BlazorUI/Components/ProgressIndicator/BitProgressIndicator.scss
+++ b/src/BlazorUI/Bit.BlazorUI/Components/ProgressIndicator/BitProgressIndicator.scss
@@ -16,6 +16,18 @@
             transform: translateX(0%);
         }
     }
+
+    @keyframes bit-pin-animation-rtl {
+        0% {
+            right: -100%;
+            transform: translateX(100%);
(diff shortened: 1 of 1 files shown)


src/BlazorUI/Bit.BlazorUI/Components/ProgressIndicator/BitProgressIndicator.scss
```diff
@@ -16,6 +16,18 @@
             transform: translateX(0%);
         }
     }
+
+    @keyframes bit-pin-animation-rtl {
+        0% {
+            right: -100%;
```
mhrastegari, line 22: You can use `inset-inline-start` in the main animation instead of `left` property

mhrastegari (changes requested):

msynk (approved):

PR #7826, opened 2024-06-21
Add missing attributes to viewport meta tag (#7784)

### LINKED ISSUE #7784: Admin Template: Menu does not open intermittent in Blazor Server Mode and IOS Safari

### Describe the bug

Only in Blazor Server RenderMode and when opening with IOS Safari:
After some time, the Burger Menu will not open the Menu or the input fields e.g. on profile page become unresonsive as well. 
The problem may arise only after some clicks and after entering some random keystrockes into a field on the profile page without the need of saving data.

When hosting as WebAssembly or Maui Blazor Hybrid RenderMode , there is no such issue. Still within IOS, when using any other Browser like Chrome, Edge, Safari and Opera, the issue does not exist.  Is is solely an Issue with IOS  & Safari & Server Rendering.

The problem only comes up with pages which have at least one EditForm with at least one Input component like the InputText. The result is the same using bit components or the standard Asp.Net Blazor InputText component.

### Expected Behavior

Menu and input fields are expected to be responsive.

### Steps To Reproduce

Setup:
- Create a new Admin Panel Template with Bit.Boilerplate 8.9.0
- Standard settings
- debug in Blazor Server mode
- open page with an IOS (ipad, iphone) Safari client (a tunnel to the hosted debugging environment may be required in order to reach the site)

Variant 1)
- Swipe to another app to hide safari
- wait a minute and and come back to safari
- try to open the menu by clicking on the burger menu

Variant 2)
- open page with an IOS (ipad, iphone) Safari client (a tunnel to the hosted debugging environment may be required in order to reach the site)
- login with default user
- navigate to the profile page
- some clicks navigation changes are required to reproduce the problem.

Variant 3)
Further tips to eventually reproduce the problem:
- enter some random data in any field
- save the data
- try to open the menu
- scroll up/down left/right
- click outside of fields
- close and reopen safari
- restart with any action a listed above

Note:
In case the issue is reproduced, it can be seen that the IOS Safari Client does not fire any Server-Action. e.g. Header.razor.cs->ToggleMenu is not called.
So it looks like that just the ISO Safari client does not react on user input and thus the call to open the Burger Menu is not performed.

### Exceptions (if any)

No exception is thrown at all.
Blazor Server:
- No Exception thrown on the Server
- The Client seems to simply not react, thus interaction with Server is not shown
- 
IOS Safari Client:
- No Exception thrown on the Client (inspected with java console Webinspector safari extension plugin does not show any error; further debugging by connecting the IOS device to a Mac has not yet been performed).

### .NET Version

8.0.301

### Anything else?

_No response_

msynk:
closes #7784

### DIFF
--- a/src/Templates/Boilerplate/Bit.Boilerplate/src/Boilerplate.Server/Components/App.razor
+++ b/src/Templates/Boilerplate/Bit.Boilerplate/src/Boilerplate.Server/Components/App.razor
@@ -14,8 +14,8 @@
 
     <meta charset="utf-8" />
     <meta name="theme-color">
-    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta name="description" content="The Boilerplate is built with ASP.NET Core, Identity, Web API, EF Core and Blazor." />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no, viewport-fit=cover" />
 
     @*#if (captcha == "reCaptcha")*@
     <link rel="preconnect" href="https://www.google.com">


ysmoradi (dismissed):

Cyrus-Sushiant (approved):

msynk:
Although this change will fix the issue on iOS, it also prevents the manual zoom feature (pinch-to-zoom) on other devices (like Android).
We are investigating it for now.

msynk:
for now I've added a check to only apply this change on iOS.

mhrastegari (approved):

msynk:
I've updated the meta content to only add the `maximum-scale=1.0`, since the other settings are not required to resolve this issue.

Issue #1970, opened 2022-05-05
Add support for Escape key press to dismiss the modal in the `BitModal` component 


msynk:
Pressing the `Escape` key on the keyboard should dismiss the BitModal.

msynk:
this is resolved in the latest improvements to BitModal

PR #8252, opened 2024-08-08
Add support for Escape key press to dismiss the modal in the BitModal component

hootanht:
Related to #1970

Add support for Escape key press to dismiss the `BitModal` component.

* **BitModal.razor**
  - Add `@onkeydown="HandleOnKeyDown"` to the root `div` element.

* **BitModal.razor.cs**
  - Add a private method `HandleOnKeyDown` to handle the Escape key press event.
  - In `HandleOnKeyDown`, check if the pressed key is "Escape" and call `CloseModal`.

* **BitModalTests.cs**
  - Add a test method `BitModalEscapeKeyPressTest` to verify the Escape key press functionality.
  - In `BitModalEscapeKeyPressTest`, simulate the Escape key press event and assert that the modal is closed.

### DIFF
--- a/src/BlazorUI/Bit.BlazorUI.Tests/Components/Surfaces/Modal/BitModalTests.cs
+++ b/src/BlazorUI/Bit.BlazorUI.Tests/Components/Surfaces/Modal/BitModalTests.cs
@@ -238,5 +238,20 @@ public class BitModalTests : BunitTestContext
         Assert.IsTrue(modalElement.ClassList.Contains(positionClass));
     }
 
+    [TestMethod]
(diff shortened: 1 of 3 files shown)


msynk:
@hootanht Thanks for submitting this PR.
but this is not the correct way to handle this issue. we need a general keyboard event handler in the BlazorUI project that can handle all the key events. implementing this new feature is not something that other people outside of the core team can handle properly. so I'm closing this PR.
**one important point**: please consider discussing your approach in the comments section of the issue first before starting to work on the PR, so we can be sure the changes are valid and accepted.

Issue #8367, opened 2024-08-21
Missing "image" in the "devcontainer.json" file


zandiarash:
After I select **"Reopen in container"** the **VScode** show an error
![22](https://github.com/user-attachments/assets/b9def849-3958-4c40-9cf9-10c3825cb067)
But when I add this line 
`"image": "mcr.microsoft.com/vscode/devcontainers/base:ubuntu",`
To my devcontainer.json file it works well.
Can I make a pull request and add this line to the projects ?

ysmoradi:
But it's working anyway, right?

ysmoradi:
But it's working anyway, right?

zandiarash:
> But it's working anyway, right?

Actually for me does not works,
Even I updated my VsCode and it's extension and dotnet.

zandiarash:
> Actually for me does not works,
> Even I updated my VsCode and it's extensions, dotnet and also fetched the latest source codes.
> Maybe I did something wrong.

Actually for me does not works,
Even I updated my VsCode and it's extensions, dotnet and also fetched the latest source codes.
Maybe I did something wrong.

zandiarash:
> But it's working anyway, right?

Actually for me does not works,
Even I updated my VsCode and it's extensions, dotnet and also fetched the latest source codes.
Maybe I did something wrong.

ysmoradi:
Could you please open this repo in github codespace?
https://github.com/bitfoundation/bit-templates-playground

zandiarash:
> Could you please open this repo in github codespace? https://github.com/bitfoundation/bit-templates-playground

That works well.

ysmoradi:
So, there's something wrong on your side, could you please compare yours with this project?

ysmoradi:
Now I see the issue you're talking about (":

zandiarash:
> Now I see the issue you're talking about (":

May you check the pull request [#8476](https://github.com/bitfoundation/bitplatform/pull/8476) ?

PR #8476, opened 2024-09-02
Image added

zandiarash:
I've Added image to fix the error of missing template
This pull request is about issue [#8367](https://github.com/bitfoundation/bitplatform/issues/8367)


ysmoradi:
I'd recommend something like this:
```json
{
    "name": "Boilerplate",
    "image": "mcr.microsoft.com/dotnet/sdk:8.0",
    "hostRequirements": {
        "cpus": 4
    },
    "onCreateCommand": "dotnet dev-certs https --trust && dotnet workload install wasm-tools",
    "waitFor": "onCreateCommand",
    "customizations": {
        "codespaces": {
            "openFiles": [
                "README.md"
            ]
        },
        "vscode": {
            "extensions": [
                "glenn2223.live-sass",
                "kevin-chatham.aspnetcorerazor-html-css-class-completion",
                "DominicVonk.vscode-resx-editor",
                "ms-dotnetools.csharp",
                "ms-dotnettools.vscode-dotnet-runtime",
                "ms-dotnettools.csdevkit"
            ]
        }
    },
    "forwardPorts": [
        4030,
        5030,
        5031
    ],
    "features": {
        "ghcr.io/devcontainers/features/node": {}
    },
    "remoteEnv": {
        "ASPNETCORE_ENVIRONMENT": "Development"
    }
}
```

ysmoradi:
The more ready image we use, the faster startup process we'll have, so instead of manually installing .NET Sdk on ubuntu, let's use .NET 8 Sdk, and instead of Ubuntu, let's use Alpine, which will be used to run backend in production by default (Unless we use something like .NET 8 SDK jimmy or something like that)

But due that fact, that image doesn't have nodejs, I've installed nodejs as a devcontainer feature, and I've also installed wasm-tools workload + C# dev kit extension

ysmoradi:
So update your PR @zandiarash

ysmoradi:
You're chaning the wrong devcontainer.json file. Revert changes and modify [this](https://github.com/bitfoundation/bitplatform/blob/develop/src/Templates/Boilerplate/Bit.Boilerplate/.devcontainer/devcontainer.json) one.

PR #11540, opened 2025-10-30
Made ReadUserSessionLogs stream logs with IAsyncEnumerable

### LINKED ISSUE #11480: Reading another users log through User management causes SignalR message size error.

- [x] I have searched the existing issues

### Describe the bug

More often then not when trying to read a remote log through user management it crashes with the message:

`Microsoft.AspNetCore.SignalR.HubException: The server closed the connection with the following error: Connection closed with an error. InvalidDataException: The maximum message size of 32768B was exceeded. The message size can be configured in AddHubOptions.`

Increasing the message size would be a temporary fix but I have gotten this error on a browser session less then 10s old.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Exceptions (if any)

_No response_

### .NET Version

_No response_

### Anything else?

_No response_

Kyemate:
closes #11480 

## Summary by CodeRabbit

* **Improvements**
  * Diagnostic logs now stream incrementally, providing faster initial response times when retrieving user session logs.
  * Added cancellation support for log retrieval operations, enabling users to stop ongoing requests.


yasmoradi:
The whole feature set will be replaced with Dev MCP soon
Thanks for your contribution

PR #12536, opened 2026-06-29
Fix memory leak issue of BitColorPicker (#640)

### LINKED ISSUE #640: BitColorPicker Memory Leak
BitColorPicker has memory leak issue when bind color value to the component and open chrome developer tools it uses lots of ram and cpu just to move color picker thumb this is caused because of JsInvokable methods continously  called in js to return mouse cursor position 
image
sometimes cpu process goes upto 100%

msynk:
closes #640

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved color picker drag behavior so pointer movement is tracked more reliably while interacting with the saturation area.
  * Reduced unintended drag handling, resulting in smoother and more accurate color selection.


msynk:
@coderabbitai re-review

msynk:
@coderabbitai re-review

PR #12550, opened 2026-06-30
Fix parsing input value issues in BitPhoneInput (#12549)

### LINKED ISSUE #12549: Parsing the input value in the BitPhoneInput is not happening on user input
Parsing the input value in the BitPhoneInput is not happening on user input:

- The component already has a correct parser, ParseFullNumber, that understands both the + and 00 international prefixes and looks up the matching country. The problem is that this parser is only wired to external changes of the full Value (via ReconcileValueAndParts).

msynk:
closes #12549

## Summary by CodeRabbit

* **New Features**
  * Phone number input now recognizes full international numbers when the value starts with `+` or `00`, and separates the local number from the detected country.
* **Bug Fixes**
  * Country detection is now applied only when it actually differs from the current selection, preventing unnecessary country updates.
  * The stored phone number now reflects the local digits rather than keeping the entire raw international string in the number field.


msynk:
@coderabbitai re-review

msynk:
@coderabbitai re-review

PR #12734, opened 2026-07-20
Correct response caching of origin-dependent and faulted responses #12733

### LINKED ISSUE #12733: fix(templates): correct response caching of origin-dependent and faulted responses
Two response-caching correctness issues in the Boilerplate template.

### 1. `GetExternalSignInUri` is shared-cacheable but its response is origin-dependent

`IdentityController.GetExternalSignInUri` is annotated with `[AppResponseCache(SharedMaxAge = 3600 * 24 * 7, MaxAge = 60 * 5)]`, and the URL it returns embeds `origin = Request.GetWebAppUrl()`.

`GetWebAppUrl()` resolves that origin from the `X-Origin` request header when it is not supplied in the query string. `AppResponseCachePolicy` varies the output cache by host, query keys and culture; `X-Origin` is in neither that vary key nor a CDN's cache key. A stored response therefore does not depend only on the inputs it is keyed by, so callers arriving from one trusted origin can be served a URL built for another.

Caching is not valuable for this endpoint, so the attribute should simply be removed rather than adding a vary rule (a CDN could not honour such a rule reliably anyway).

### 2. A response refused by the output cache keeps its shared `Cache-Control`

`AppResponseCachePolicy.CacheRequestAsync` runs before the endpoint and writes `Cache-Control` with the configured shared TTLs. `ServeResponseAsync` later sets `AllowCacheStorage = false` for non-200 responses and for responses carrying a non-culture `Set-Cookie`, but it never revises that header.

The result is that a response ASP.NET Core's output cache has just declined to store is still advertised to CDNs and browsers as publicly cacheable. For example a 404 from `/api/v1/ProductView/Get/{id}` goes out with `public, s-maxage=604800`, so creating that product a minute later leaves the stale 404 on the edge for up to seven days.

This also means the comment in `WebServerExceptionHandler` stating that setting the status code "ensures that a faulted response is not cached" only holds for the output cache, not for the edge or the browser.

### Notes

Both paths are reachable only when `ResponseCaching:EnableOutputCaching` / `ResponseCaching:EnableCdnEdgeCaching` are enabled; both default to `false` in the template's `appsettings.json`.

yasmoradi:
Closes #12733

### 1. `GetExternalSignInUri` is no longer shared-cacheable

The URL this action returns embeds `origin = Request.GetWebAppUrl()`, and `GetWebAppUrl()` falls back to the `X-Origin` request header when the origin is not in the query string. `AppResponseCachePolicy` varies by host, query keys and culture, so `X-Origin` is in neither the output cache's vary key nor a CDN's cache key, and a stored response did not depend only on the inputs it was keyed by.

The `AppResponseCache` attribute is simply removed. Caching buys little here, and a vary rule would not be a real fix since CDNs do not honour `Vary` on arbitrary headers reliably.

### 2. A response the output cache refuses to store no longer advertises shared caching

`CacheRequestAsync` writes `Cache-Control` before the endpoint runs. `ServeResponseAsync` already set `AllowCacheStorage = false` for non-200 responses and for responses carrying a non-culture `Set-Cookie`, but left that header untouched, so CDNs and browsers kept caching a response ASP.NET Core had just declined to store. A 404 from `/api/v1/ProductView/Get/{id}` went out with `public, s-maxage=604800`, leaving a stale 404 on the edge for up to seven days after the product was created.

The header is now downgraded to `no-store, private` alongside `AllowCacheStorage`.

The `response.HasStarted is false` guard matters: `ServeResponseAsync` also runs for streamed responses, whose headers are already flushed and read-only — without it the middleware throws `InvalidOperationException: Headers are read-only, response has already started`. That path needs no downgrade anyway, since enabling shared caching is exactly what suppresses streaming pre-rendering (see `HttpRequestExtensions.IsStreamPrerenderingSuppressed`).

### Also included

`IdentityController.WebAuthn` assertion options:

- The cache key was built as `new string(challenge.Select(b => (char)b))`, putting NUL and control characters into the Redis key. It is now base64url encoded.
- The options were deliberately left in the cache until they expired, because the two-factor round trip verifies the same client response a second time. They are now evicted once the flow reaches its final step, so a WebAuthn challenge stops being reusable for the remainder of its 3 minute window. `SignIn` throws on an invalid two-factor code, so a failed attempt still leaves the options in place for a retry.

### Verification

- `dotnet build` on `Boilerplate.Server.Web` — succeeds, no new warnings.
- `dotnet test --filter "TestCategory=Caching"` — passes.
- `dotnet test --filter "TestCategory=SEO"` — 3/3 pass.

`Prerendering_WithOutputCaching_Should_ReturnCompleteNonStreamedHomePage` is what caught the missing `HasStarted` guard.

## Summary by CodeRabbit

- **Bug Fixes**
  - External sign-in URLs are now generated per request and no longer reused from cache.
  - Output caching is better isolated by tenant and request path, and non-cacheable responses are reliably marked private and not stored.
  - WebAuthn two-factor verification now keeps intermediate state correctly and applies credential counter changes only on completion.
  - Culture prefixes in URLs are removed only when they appear at the start of the path.

### DIFF
--- a/src/Templates/Boilerplate/Bit.Boilerplate/.docs/14- Response Caching System.md
+++ b/src/Templates/Boilerplate/Bit.Boilerplate/.docs/14- Response Caching System.md
@@ -150,18 +150,29 @@ Output cache still works correctly for multi-language scenarios.
 public async ValueTask CacheRequestAsync(OutputCacheContext context, CancellationToken cancellation)
 {
     var responseCacheAtt = context.HttpContext.GetResponseCacheAttribute();
-    
+
     if (responseCacheAtt is null) return;
 
-    // Default: SharedMaxAge = MaxAge if not specified
-    if (responseCacheAtt.SharedMaxAge == -1)
-    {
-        responseCacheAtt.SharedMaxAge = responseCacheAtt.MaxAge;
-    }
+    context.AllowLocking = true;
+    context.EnableOutputCaching = true;
 
-    var clientCacheTtl = responseCacheAtt.MaxAge;      // In-memory + Browser
-    var edgeCacheTtl = responseCacheAtt.SharedMaxAge;  // CDN Edge
-    var outputCacheTtl = responseCacheAtt.SharedMaxAge; // ASP.NET Core Output Cache
+    // What the output cache keys on, besides the request path:
+    context.CacheVaryByRules.QueryKeys = "*";
+    context.CacheVaryByRules.VaryByHost = true;
+    context.CacheVaryByRules.HeaderNames = new[] { HeaderNames.Origin, "X-Origin" };
+    context.CacheVaryByRules.VaryByValues.Add("Culture", CultureInfo.CurrentUICulture.Name);
+
+    // Multi-tenant: an authenticated request resolves its tenant from the user's claim rather than from the
+    // host, and tenant scoped entities are filtered by it, so two tenants on one host must not share an entry.
+    if (context.HttpContext.User.GetTenantId() is Guid currentTenantId)
+        context.CacheVaryByRules.VaryByValues.Add("Tenant", currentTenantId.ToString());
+
+    // SharedMaxAge falls back to MaxAge when it isn't set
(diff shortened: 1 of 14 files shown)


yasmoradi:
@coderabbitai review

PR #13129, opened 2026-09-02
Apply Butil improvements I (#13086)

### LINKED ISSUE #13086: `Butil` improvements I
`Butil` improvements I:

### Web Workers
`Worker`, `SharedWorker`, `MessageChannel` / `MessagePort`. There is no CPU-offload path in the
package at all. `ServiceWorker` and `BroadcastChannel` are wrapped, but dedicated and shared workers
are absent, and with them the whole structured-messaging surface. The design question worth settling
early: what does a worker *run*? Options are a user-supplied script URL, a Butil-provided shim that
marshals to a second .NET runtime instance, or both.

### WebSocket
`EventSource` (server-sent events) is wrapped; WebSocket is not. On WebAssembly `ClientWebSocket`
is emulated over the JS `WebSocket` anyway, so a first-class wrapper costs little and gives back
binary frames, close codes, `bufferedAmount` and sub-protocol negotiation.

### Canvas 2D, OffscreenCanvas, WebGL, WebGPU
No drawing surface of any kind. Also missing as a consequence: `toBlob` and `toDataURL`, which means
there is no way to capture a video frame, export a generated image, or produce a thumbnail from C#.
A minimal, high-value first slice is `toBlob` / `toDataURL` plus `drawImage` from a video element -
that alone covers screenshotting and thumbnailing without committing to a full 2D context API.

### WebRTC
`RTCPeerConnection`, `RTCDataChannel`, `getStats`. `MediaDevices` acquires the stream and
`MediaRecorder` writes it to a file, but nothing sends it to another peer.

### Streams API
`ReadableStream`, `WritableStream`, `TransformStream`, plus `tee` and `pipeThrough` / `pipeTo`.
The package already uses streams internally - `Scripts/fetch.ts` reads a response body through
`getReader` for progress, and `Scripts/compression.ts` pipes through the native codec - but never
exposes them, so a consumer cannot build a pipeline of their own or hand a stream between APIs.

### requestAnimationFrame, requestIdleCallback, and the Scheduler API
`scheduler.postTask`, `scheduler.yield`, and `navigator.scheduling.isInputPending`. There is no
frame loop or task-priority control available to C#. `requestAnimationFrame` is used internally by
the gamepad and barcode-detector modules, so the interop shape is already proven.

### window.postMessage and iframe messaging
`postMessage` appears only in the BroadcastChannel and ServiceWorker modules. There is no
cross-frame or cross-origin window messaging - no way to talk to an embedded iframe, or to a window
opened with `Window.Open`.

### DOM creation and querying
Every DOM API in the package is anchored to an `ElementReference` that Blazor already rendered.
Missing: `createElement`, `getElementById`, `querySelector` returning a usable handle, and node
traversal (`parentNode`, `children`, `nextElementSibling`). This is a deliberate design boundary
today, but it blocks any scenario where the element is not one of the app's own rendered elements.

### Shadow DOM
`HasShadowRoot` exists as a probe. Missing: `attachShadow`, access to a shadow root, and querying
into one. Relevant for interoperating with third-party web components.

### DataTransfer and drag-and-drop payloads
Drag event args are wrapped, but the payload is not: `files`, `getData` / `setData`, `dropEffect`
and `setDragImage` are all unreachable. Handling a file dropped onto the page currently needs
hand-written JS.

### getComputedStyle and CSSOM
Style access is inline-only, which the XML docs state explicitly. Missing: `getComputedStyle`,
`CSS.supports`, stylesheet manipulation (`insertRule` / `deleteRule`), `CSS.registerProperty`, and
the CSS Custom Highlight API.

### AbortController and AbortSignal as first-class types
Abort handling is baked into `Fetch`, `Nfc` and `IdleDetector` individually. There is no standalone
controller a consumer can create, pass around, and use to cancel several operations at once, and no
`AbortSignal.timeout`.

msynk:
closes #13086

## Summary by CodeRabbit

* **New Features**
  * Added browser interoperability for DOM, CSS, Canvas, Shadow DOM, drag-and-drop, cancellation, streams, scheduling, workers, messaging, WebRTC, and WebSockets.
  * Added canvas export, streaming pipelines, task scheduling, transferable messages, peer connections, and socket communication.
  * Added interactive demo and sample pages for the new capabilities.
* **Documentation**
  * Expanded API documentation and navigation.
* **Tests**
  * Added end-to-end coverage for the new browser features and communication scenarios.


msynk:
@coderabbitai full-review

msynk:
@coderabbitai full-review

PR #13210, opened 2026-09-12
Fix focus issues of overlay components (#13192)

### LINKED ISSUE #13192: Tap away when focus is on entry field does not pick up new value

- [x] I have searched the existing issues

### Describe the bug

 Version: Bit.BlazorUI 10.6.0-pre-04 (regressed from 10.5).
  - Repro: a non-blocking BitModal containing <BitTextField @bind-Value="Model.Name" />. Type into the field, then click the overlay without leaving the field. OnOverlayClick / OnDismiss fire, but the bound property still holds the old value — the change event never arrives. Clicking any button inside the modal first commits the value
    normally.
  
I did not test this beyond my text fields.  In my case they can change the value of text field and when they tap away, I check to see if the view model has changed and then save their work.

### Expected Behavior

- Expected (10.5 behaviour): the focused input blurs and commits before the dismiss handlers run.

### Steps To Reproduce

_No response_

### Exceptions (if any)

_No response_

### .NET Version

_No response_

### Anything else?

_No response_

msynk:
closes #13192

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved focus handling for dialogs, modals, panels, and overlays.
  - Overlay interactions now commit active input values before dismissal.
  - Focus remains within active surfaces and is restored when overlays close.
  - Improved Tab and Shift+Tab keyboard navigation.

- **Accessibility**
  - Consumer-provided tab index values are now respected.
  - Programmatic focus avoids unnecessary outlines while preserving keyboard navigation indicators.

- **Demo**
  - The dialog example now displays the currently entered field value.

### DIFF
--- a/src/BlazorUI/Bit.BlazorUI/Components/Surfaces/Dialog/BitDialog.razor
+++ b/src/BlazorUI/Bit.BlazorUI/Components/Surfaces/Dialog/BitDialog.razor
@@ -6,7 +6,22 @@
     @* While a kept-mounted Dialog is closed it is still in the page, only hidden: inert and aria-hidden
        take it out of the tab sequence and out of the reading order for the browsers that would otherwise
        still reach into a display:none subtree through a stale reference. *@
+    @* The keydown handler sits on the root so that Escape dismisses the Dialog from anywhere inside it -
+       the surface, the overlay, whichever part of it the keyboard happens to be in. The keys stop here:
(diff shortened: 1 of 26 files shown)


msynk:
@coderabbitai full-review

msynk:
@coderabbitai full-review

PR #13262, opened 2026-09-17
Add Bit.Minifier to shrink published Blazor WebAssembly assemblies #13261

### LINKED ISSUE #13261: Blazor WebAssembly publishes ship trimmed assemblies that nothing minifies
A Blazor WebAssembly publish runs ILLink, which removes the .NET code an app doesn't use. What it leaves is not minified. The JavaScript and CSS of the bit packages already go through esbuild and sass, but the `.wasm` assemblies still carry:

- thousands of attributes only the compiler reads (`[CompilerGenerated]` on accessors, `[Nullable]`, `[NullableContext]`, `[NotNullWhen]`, ...). In the trimmed Bit.BlazorUI of a small app they fill 8,779 of the 10,588 CustomAttribute rows;
- long compiler-generated names such as `<BuildRenderTree>b__12_0` and `<OnClickAsync>d__31`, about 75K characters in the same assembly;
- private and internal names nothing reads at runtime.

No existing tool removes this and keeps the result debuggable:

- trimming and `WasmStripILAfterAOT` only remove code;
- obfuscators rename everything, which breaks `[Parameter]`, `[JSInvokable]` and JSON names, and they don't keep portable pdbs in sync.

A publish-time step after ILLink can remove it. It has to keep the portable pdbs matching, keep public names so logged stack traces stay readable, and leave alone anything the runtime, Blazor or serializers read by name. Measured on bitplatform.dev's WebAssembly client (69 managed `.wasm` files, 2,094 KB brotli):

| | brotli |
|---|---:|
| attributes and generated names only, keeping what other libraries read | -68 KB (-3.2%) |
| every non-public name, parameter name and non-public property too | -185 KB (-8.8%) |

The ceiling is low because IL itself is already compact, and because local variable names live only in the pdb, which isn't downloaded.

yasmoradi:
closes #13261

Adds **Bit.Minifier**, a new package. Installed in a Blazor WebAssembly project, it runs right after ILLink on `dotnet publish` and rewrites every trimmed assembly in `obj/` together with its portable pdb. The only setting is `BitMinifyAggressive`. The build output, `dotnet run` and `dotnet watch` are never touched. Details: [src/Minifier/README.md](https://github.com/yasmoradi/bitplatform/blob/13261/src/Minifier/README.md).

## What it does

**Default level**
- **Attributes:** removes the ones only the compiler reads: `[CompilerGenerated]` on methods, the nullable metadata, the `System.Diagnostics.CodeAnalysis` nullability hints, and BlazorUI's source generator markers.
- **Generated names:** `<BuildRenderTree>b__12_0` becomes `<a>b__12_0`. The `<X>kind__` shape debuggers parse is kept.
- **What other libraries read stays:** backing-field names and `[CompilerGenerated]` on fields (EF Core, Newtonsoft.Json), and the nullable metadata whenever EF Core is in the app. The bit libraries, which read none of these, lose them too.

**`BitMinifyAggressive=true`**
- **Everything else:** also renames every non-public type, method and field in every assembly, clears non-public parameter names, removes non-public property definitions, and drops the attributes only debuggers and analyzers read.
- **Public names stay**, so logged stack traces still name the public frames.
- **Also kept:**
  - names that a string literal mentions (`nameof`, `GetMethod("...")`, `[UnsafeAccessor(Name = "...")]`); this also keeps the backing fields EF Core's compiled models reach by name;
  - public properties and fields, even of internal types, which System.Text.Json serializes;
  - constructor parameters;
  - members with runtime attributes;
  - Blazor component types, whose names the prerendering server sends to the client;
  - types an embedded resource is named after;
  - JS interop's generated registration and `<Main>$`;
  - every name in `System.Private.CoreLib` and `System.Runtime.InteropServices.JavaScript`.
- **Not protected at this level:** libraries that read backing fields or nullable metadata by themselves, such as EF Core without a compiled model. The legacy chart of the BlazorUI demo, which uses Newtonsoft.Json, works at this level.

**Also**
- **Nullable metadata** is kept when the app sets `NullabilityInfoContextSupport` to `true`.

**Safety**
- **References checked first:** nothing is written unless every reference into the rewritten assemblies still resolves. Otherwise the publish gets warning `BITMIN001` and ships ILLink's output.
- **Only changed assemblies are rewritten**, and the output is byte-identical between publishes, wherever the project lives.
- **Name map:** `obj/.../bit-minifier.map` lists every renamed member, for reading stack traces.

## Numbers

bitplatform.dev's WebAssembly client, published in Release: 69 managed `.wasm` files, 2,094 KB brotli (6,755 KB raw).

| | brotli | raw |
|---|---:|---:|
| default | 2,026 KB (-68 KB, -3.2%) | 6,403 KB (-5.2%) |
| aggressive | 1,909 KB (-185 KB, -8.8%) | 6,023 KB (-10.8%) |

A test app using about 15 components of BlazorUI, Extras, Butil, Bswup and Brouter went from 1,953 KB to 1,883 KB (default) and to 1,779 KB (aggressive). The tool takes about 3 s per publish.

## Tested

- **Unit tests:** 21 tests (`src/Minifier/Tests`) over two fixture assemblies. They cover:
  - attributes, names, and behaviour before and after;
  - stack-trace line numbers and pdb consistency (id, sequence points, locals, state machines);
  - friend assemblies, the rules for other libraries and EF Core detection;
  - whole-folder mode, and the aggressive level, including members looked up through `nameof` and backing fields reached through `[UnsafeAccessor]` and `GetField`;
  - the refuse-to-write path and idempotence.
- **Browser, test app:** published from packages and project references at both levels. Text field, button, dropdown, modal, date picker, pivot, Brouter navigation and data grid all work with no console errors.
- **Browser, bitplatform.dev:** published at both levels. The same 8 pages render as without the minifier, with no console errors. Two runtime bindings were found and protected along the way: `<Main>$` and JS interop's `__GeneratedInitializer`.
- **Packaging:** incremental publishes skip the step, and the package installs from a local feed.
- **Deployed at the aggressive level:** the BlazorUI demo's legacy chart (Newtonsoft.Json) works. The Boilerplate's offline todo (EF Core compiled model) broke: its backing fields were renamed although the compiled model names them in strings. That is fixed and covered by a test, at no size cost on bitplatform.dev (the same names are renamed, and it is 0.5 KB smaller), but it hasn't been redeployed yet.

## Repo changes

- **New:** `src/Minifier` (the tool, the `buildTransitive` targets, tests, README), `Bit.Minifier.slnx`, and `bit.ci.Minifier.yml`.
- **Solutions:** `Bit.slnx` and `Bit.CI.Release.slnx` include the project.
- **Release and CI:** `nuget.org.yml`, `prerelease.nuget.org.yml` and `all.ci.yml` build and pack it the way they do `Bit.CodeAnalyzers`, and `all.ci.yml` also runs its tests.
- **CI:** not run on the fork yet.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


yasmoradi:
Two gaps in how `BitMinifySuperAggressive` / `BitMinifyAggressive` find names that are read at runtime.

`AggressiveMinifier.Collect` builds one word set for the whole app: every `*.dll` in ILLink's output folder, plus the unminified assemblies in super aggressive mode. So a `GetMethod("Foo")` in one assembly still protects `Foo` in another. That set only has three sources: `ldstr` operands, attribute string arguments and resource **names**. Two places where a name can appear are never read:

1. **Contents of embedded resources.** Only `resource.Name` is added. An embedded `appsettings.json` (or any other text resource) that names a type, such as `Logging:LogLevel` categories or a type name in custom config, doesn't protect it, and super aggressive renames it.
2. **UTF-8 string literals (`"..."u8`).** The compiler stores these as RVA byte data in `<PrivateImplementationDetails>` and loads them with `ldsflda`, not `ldstr`, so none of their words are collected.

Neither one affects bitplatform.dev today: its embedded `appsettings.json` only holds `ApiServerAddress`, and neither the site nor Bit.BlazorUI uses `u8` literals. Other apps can hit both.

Possible fix: in `Collect`, run `AddWords` on the decoded text of `EmbeddedResource`s that look like text (UTF-8 without invalid sequences, e.g. `.json`/`.xml`/`.txt`, skipping `.resources`, which already keep the type by name), and on the `InitialValue` of the RVA fields that `ldsflda` loads next to a `ReadOnlySpan<byte>` constructor. Add a test for each.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

yasmoradi:
### Benchmark: what the Platform website actually downloads

Published `Bit.Websites.Platform.Server` four times from this branch's tree, changing only the minifier (`obj/Release` + `bin/Release` of the client wiped each time so ILLink re-ran on untouched assemblies), then loaded `/contact-us` with a cold cache — 99 same-origin files, no third-party requests (the home page was avoided, it pulls YouTube). The server negotiates brotli, so every size below is the `.br` on the wire.

29 of those files are byte-identical across all four publishes — every css/js/image, `manifest.json`, `favicon.ico`, plus `dotnet.native.wasm`, `dotnet.runtime.js`, `dotnet.native.js` — 755,510 B in total. They are excluded below. What changes is the 69 managed assemblies plus `_framework/dotnet.js`:

| config | the 70 changed files | vs no minifier | `Bit.BlazorUI.wasm` |
| --- | --- | --- | --- |
| no minifier | 2,217,862 B | — | 330,470 B |
| aggressive @ `4e38cea` | 2,023,883 B | −8.75% | 291,771 B |
| aggressive @ `a3148a1` | 2,029,384 B | −8.50% | 291,629 B |
| super aggressive @ `a3148a1` | 2,003,755 B | −9.65% | 284,879 B |

Whole page: 2,973,372 → 2,779,393 / 2,784,894 / 2,759,265 B.

**Finding: plain aggressive lost 5,501 B between `4e38cea` and `a3148a1`.** Looking at only `Bit.BlazorUI.wasm` hides it (that one got 142 B smaller); across the whole download set aggressive is now slightly worse than before the super-aggressive work landed. Biggest regressions: `System.Text.Json` +448, `Bit.Websites.Platform.Client` +441, `Microsoft.AspNetCore.Components` +394, `System.Net.Http` +319, `System.ComponentModel.TypeConverter` +313 — partly offset by `System.Text.RegularExpressions` −383 and `System.Private.CoreLib` −355. Looks like aggressive now keeps some names it used to drop; worth a check that this was intended.

Super aggressive is unaffected: it takes another 25,629 B (−1.26%) off aggressive @ `a3148a1`, mostly `Bit.BlazorUI` (−6,750), `System.Text.Json` (−2,053), `Microsoft.AspNetCore.Components` (−1,651) and `Bit.Websites.Platform.Client` (−1,590). `System.Private.CoreLib` barely moves between aggressive and super (−1,059 B, −0.2%).

Sanity check: the site was opened and the theme toggle exercised in all four publishes — works everywhere, no console errors.

msynk (approved):

